PR #1, opened 2015-11-05
Fix typo in index.js: routeState to routing

ellbee:
Thanks for creating this, it is really neat!  I was playing around with using it with redux-devtools and came across a runtime error, which I believe to be a simple typo.


jlongster:
Oh man! You are right. I hadn't tested the store->router changes after the latest refactor. Thanks a lot!

Let me know if something else breaks with the devtools. It should work ok.

ellbee:
It seems to be working great!

PR #3, opened 2015-11-06
Typo

grabbou:
This is rather funny and unnecessary pull request, but wanted to say thanks for the idea, gonna try it in next project :+1: 

Feel free to close it and fix directly, or just leave it that way, comments are not an issue at all

PS. Not a native guy, but to me `when` might be a better fit then `if` #grammar


jlongster:
Works for me, I think `if` is fine here but I appreciate the typo fix! Let me know how it works out, I extracted it from a local small project of mine so it's likely there are improvements.

PR #4, opened 2015-11-06
Fix typos

ingro:


jlongster:
Ah, thanks! Renamed the project last minute.

PR #6, opened 2015-11-06
Restore location after devtools reset

ellbee:
I found a problem with Devtools where a 'RESET' will change the path to '/' regardless of the initial URL. This is the quick workaround I am currently using, but I wonder if there is a better way.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -15,7 +15,10 @@ function updatePath(path, noRouterUpdate) {
 
 // Reducer
 
-const initialState = {};
+const initialState = typeof window === undefined ? {} : {
+  path: locationToString(window.location)
+};
+
 function update(state=initialState, action) {
   if(action.type === UPDATE_PATH) {
     return Object.assign({}, state, {


jlongster:
Good catch! I will test it out myself as well and think about it. I'd like if it we didn't have to explicitly handle this, but we'll see. Might take me a day or two to look at this more deeply.

ellbee:
Thanks, that sounds good, it would definitely be preferable to not have this code in the project!

Actually, my initial thought with this was just to set the initialState of routerReducer.path to locationString(window.location) when running in a browser. For some reason that eludes me now I decided it wouldn't work, but having thought about it some more it seems like it would work OK:

```js
const initialState = typeof window === undefined ? {} : {
  path: locationToString(window.location),
};
```

jlongster:
@ellbee oh, good point! I don't see any problem with that. Can you test it and see if it works, and possibly update this PR? If not I can get to it this week.

ellbee:
Sure, I'll have some time to look at it later today.

ellbee:
I've updated the PR. It is working fine in the tests I have done.

jlongster:
Looks good, thanks!

PR #7, opened 2015-11-06
Eliminate double ? from routing.path

ellbee:
I was getting double ? before the query string in routing.path. I think this is because the location.query already has a ? in the front (and the same applies for hash). I removed the '?' that was being prepended which seems to fix the problem unless there are some edge cases I am missing.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -29,9 +29,7 @@ function update(state=initialState, action) {
 // Syncing
 
 function locationToString(location) {
-  return location.pathname +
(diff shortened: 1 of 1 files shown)


jlongster:
It seems you are right. Not sure why I did that, must have been from some previous code that is invalid now.

I'd like to write tests soon, but just wanted to get this out the door to see what people think. Thanks.

ellbee:
Cool. I appreciate you getting it out of the door, it has certainly scratched an itch for me.

PR #11, opened 2015-11-08
ES2015 improvements

StevenLangbroek:
This cleans up the code a bit:

- The action creator can use implicit return wrapped in parentheses, and now implements enhanced object literals more thoroughly
- Rewrite the reducer to use a ternary, as this will only ever have one `action.type` as far as I can tell...
- Declare everything as a const for consistency

I understand this is all personal preference, so feel absolutely free to decline this PR :).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,34 +5,30 @@ const UPDATE_PATH = "@@router/UPDATE_PATH";
 
 // Action creator
 
-function updatePath(path, noRouterUpdate) {
-  return {
-    type: UPDATE_PATH,
-    path: path,
-    noRouterUpdate: noRouterUpdate
-  }
-}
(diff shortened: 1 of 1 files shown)


jlongster:
Thank you very much for the PR, but as this is just preference like you mentioned, I'm going to politely decline. I disagree with the movement to never use `function`, and defining it as `const` actually will hurt live-reloading in the future because devtools will not be able to hot swap functions. When we get native ES6 modules debuggers should be able to eval inside modules and you can just interactively re-define functions for REPL-style development. It's also a lot more syntactically noisy (arrow functions are much more succinct for anon functions, but not as much for defining functions).

I hope this doesn't deter you from future PRs, I'm sure there are other things that should be fixed here. Thanks again!

PR #15, opened 2015-11-10
Allow users to provide custom selectRouterState

davezuko:
I have no idea if this is desirable or not, but figured I'd give it a shot just in case. I have a few projects with redux-router that are candidates for conversion to redux-simple-router, and it would be nice to be able to customize what the name of the property is on the store to maintain consistency.

If this is in fact something that might make sense but you can think of a more elegant implementation, I'm all ears. If not, I understand.

Thanks for a cool project.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@
 // constants
 
 const UPDATE_PATH = "@@router/UPDATE_PATH";
+const SELECT_STATE = (state) => state.routing;
 
 // Action creator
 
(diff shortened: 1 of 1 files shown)


gaearon:
Better to let user provide `selectRouterState` with default being `state => state.routing`.

davezuko:
@gaearon I updated the PR, does this look closer to what you'd imagine? We could always add an invariant in here as well to make sure that the optional `selectRouterState` is in fact a function and then provide a better error message, but I wanted to make sure this is even desirable first.

jlongster:
Yeah this is a great idea, I was going to get around to it. Thanks! I'll take a look and merge it soon.

crossman:
Anything I can do to help get this merged in? This would be really useful. Documentation?

jlongster:
Been slammed at work the past few days, will definitely look tomorrow though! And yes, we do need to document it but shouldn't be much so I can add it in.

jlongster:
Thanks for this, I'll write some docs when I publish which should be today/tomorrow.

jlongster:
Oops, do you mind rebasing? I merged some other PRs first.

jlongster:
I'll cut a new release once this gets merged.

davezuko:
Hey, currently flying but I'll rebase this tonight once I get to my hotel. Thanks!

davezuko:
@jlongster rebased, let me know how this looks.

jlongster:
Thanks!

PR #18, opened 2015-11-11
Update syncReduxAndRouter to return an unsubscribe function

tappleby:
Just in case someone wants to clean up the listeners.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -43,15 +43,15 @@ function syncReduxAndRouter(history, store) {
     );
   }
 
-  history.listen(location => {
+  const unsubscribeHistory = history.listen(location => {
     // Avoid dispatching an action if the store is already up-to-date,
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -62,6 +62,11 @@ function syncReduxAndRouter(history, store) {
       history.pushState(null, routing.path);
     }
   });
+
+  return function unsubscribe() {
```
tappleby: @jlongster Do you think we would ever want to return anything else from syncReduxAndRouter? If so this could return an `{ unsubscribe }` shape instead.

jlongster: I don't think so. I think just returning a function is fine! Worst case, this should only ever be used once in a project, so it wouldn't be a big deal to do a backwards-incompatible change.

jlongster:
Thanks!

PR #19, opened 2015-11-11
Replace es2015 Object.assign for non es2015 environments

kmudrick:
When I tried using this library in a node 0.12.7 environment, my server side code failed due to Object.assign not being present.  Since the source is already being transpiled with es2016 presets, it seems to make sense to also replace Object.assign when not available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   ],
   "license": "MIT",
   "scripts": {
-    "build": "mkdir -p lib && babel ./src/index.js --presets babel-preset-es2015 --out-file ./lib/index.js",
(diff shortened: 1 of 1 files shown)


jameswomack:
:+1:

kmudrick:
Any chance this can get merged in?  I have to shim in Object.assign via https://www.npmjs.com/package/object.assign otherwise

jlongster:
Sorry! Been really busy at work, but was going to merge all the PRs today. I'll do it now.

kmudrick:
Thanks!

dtschust:
Is there an ETA on this fix being published on npm? No problem if it's going to be a while, I can just point to github directly.

PR #23, opened 2015-11-12
fix locationToString didn't match createHashHistory routing path

zeroshine:


jlongster:
Sorry I haven't looked at this more yet. I'll definitely look into it this week.

jlongster:
Thanks for this PR. I'm going to try to look for a more robust solution though. This code is not obvious and you also are dropping the query params when not using hash history.

PR #27, opened 2015-11-14
add setTimeout to syncReduxAndRouter's store listener so we go last

scottnonnenberg:
First of all, thank you so much for this great example of minimalism. Playing around with this yesterday and today, I ran into an issue where, on a transition caused by changing the path in state directly, I would see the target views rendered twice. More importantly, the first time they would fail, and the second time they would succeed.

The issue is this: `react-redux`'s higher-order React components installed via `connect()` only subscribe to state changes on `componentDidMount`, which means that they are guaranteed to be _after_ the listener installed by `redux-simple-router`'s `syncReduxAndRouter`. And that means we'll fire a history transition before any React components have the new data. And if those React components assume certain data will be there, you'll get a crash, like I did.

This change introduces a `setTimeout` into the state listener installed by `syncReduxAndRouter`, to ensure that it runs only after all the other store listeners have run.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -52,15 +52,19 @@ function syncReduxAndRouter(history, store) {
   });
 
   const unsubscribeStore = store.subscribe(() => {
-    const routing = store.getState().routing;
-    // Don't update the router if nothing has changed. The
-    // `noRouterUpdate` flag can be set to avoid updating altogether,
-    // which is useful for things like loading snapshots or very special
-    // edge cases.
-    if(routing.path !== locationToString(window.location) &&
-       !routing.noRouterUpdate) {
-      history.pushState(null, routing.path);
-    }
+    // Wait until all other listeners have fired and all react components are updated
+    // before triggering the new route; this ensures they all have the new data first.
+    setTimeout(function() {
+      const routing = store.getState().routing;
+      // Don't update the router if nothing has changed. The
+      // `noRouterUpdate` flag can be set to avoid updating altogether,
+      // which is useful for things like loading snapshots or very special
+      // edge cases.
+      if(routing.path !== locationToString(window.location) &&
+         !routing.noRouterUpdate) {
+        history.pushState(null, routing.path);
+      }
+    }, 0);
   });
 
   return function unsubscribe() {


jlongster:
Sorry I haven't gotten to this yet. I'm reluctant to do this because adding `setTimeout` makes code harder to follow. Can you show me a test case that fails because of this? How are you using `connect`, what state are you listening to, and what triggers the route change?

You shouldn't normally be dependent on the order of subscribers.

scottnonnenberg:
Given that there's no test harness, no I don't have a failing test ready for you. :0)

Without throwing a lot of code on here, I think I can just describe it like this: given an action that changes both data and the router path, do you want to transition the URL first, rendering them, and then alert all the views to the new data, or vice versa? Store subscriber order definitely matters here, and by default the URL transition will happen first. Given that, without this `setTimeout` change, the solution is to harden all views to the data they'd have both before and after any url-changing redux action.

So it could just be a matter of taste: I prefer that the new data hit all the views, and _then_ we change the URL, rendering all the new views at that new URL.

While we're on the topic of taste, in my exploration yesterday I also found a `debounce` useful for the `history.listen()` callback, so a redirect away from a URL in an `onEnter` doesn't result in two urls in your redux state, just the target of the redirect.

jlongster:
> given an action that changes both data and the router path, do you want to transition the URL first, rendering them, and then alert all the views to the new data, or vice versa?

I think this is confusing because of different assumptions. You should not be "selecting" on the URL in the redux app state. You should be using react-router to automatically render the new routing components when routes change. If you do that I don't really think you have this problem.

Let's go through what happens with the scenario you mentioned:

1.The `UPDATE_PATH` action is fired and the path updates, and you also update some other data in your app state (this is the only way 1 action changes both data and path like you said)
2. The `syncReduxAndRouter` listener is going to fire first, and a `history.pushState` call happens. This (I think) renders the new routing components synchronously.
3. Any "connected" components are also rerendered with the new app state. But note that these might even be old routing components for the previous URL, but that's fine because a `setState` for those are no-ops (if they aren't mounted).

The only way to update both data and router path at the same time means that the new routing components *will* be rendered with the new app state, because the `history.pushState` won't happen until the new app state is installed.

If you can give a realistic use case that is hard with the current setup, I'm happy to talk more about this. But currently I don't see what it solves and these kinds of things make the user be aware of internals.

scottnonnenberg:
I appreciate the time you're taking to walk through this with me. I'm definitely not doing anything funky with the router state - I'm fully relying on `react-router` to do the right thing. And I agree that most of the time you should be using standard router calls, standard `Link` components, etc. to change router state.

But here's my real scenario. Let's get concrete. The user is logging in, and right after they log in, they need to go to their profile page. Or perhaps even the page they were originally trying to go, saved on redux state. The cleanest way I can find for this kind of situation is a reducer which looks for the 'login succeeded' action, then updates the router's path in redux. And because this is the 'login succeeded' action, it also carries the current user information.

In this situation, `react-simple-router` results in an attempt to render the user's profile page before the new redux state, with the user data now added, hits the views. Similar to what you mention above, but no `UPDATE_PATH` action - it's the other direction. The profile page is immediately re-rendered as more store `subscribe()` listeners are called with the new user-containing state. To the user, this can manifest as an ephemeral crash, or perhaps just a visual flash. You could very easily solve it with some sort of spinner.

I suppose my point is that this is unexpected behavior. And it would be solved with a `setTimeout` call. That's it.

jlongster:
Thanks for the reply!

> The cleanest way I can find for this kind of situation is a reducer which looks for the 'login succeeded' action, then updates the router's path in redux. And because this is the 'login succeeded' action, it also carries the current user information.

How does a reducer update the path? That should be impossible. You can only update the path by firing an action (or in a routing component, accessing `this.props.history`.

I think there is another workflow for what you want to do that does not suffer from this. I know others have implemented similar things, and you might want to check out https://github.com/erikras/react-redux-universal-hot-example/ in a week or so because we are porting it to this project and it'll be an example of a lot of complex workflows.

Here's a great use case why I'm reluctant to do this: write now tests are super easy to write because everything is synchronous: https://github.com/jlongster/redux-simple-router/blob/master/test/index.js. Making tests async is a big burden. In fact the only way I could test this is to do more `setTimeout`s.

`dispatch(updatePath('/foo'))` will synchronously update the path. Whereever you are updating the user info, maybe a `dispatch(updateUser(info))` action, just make sure all that happens before the route is updated.

scottnonnenberg:
Yep, looks like what I was doing is now impossible. This was the line that enabled changing the state in a random reducer and having the URL change: https://github.com/jlongster/redux-simple-router/blob/d9087413fa741ab33df3e4c0753bc4c6c3bfa814/src/index.js#L64

PR #34, opened 2015-11-18
Add replacePath functionality

MattKunze:
Pretty much the same as updatePath except it replaces the current
path in history rather than appending a new entry

Fixes jlongster/redux-simple-router#32


jlongster:
Sorry for the duplicate work! Let's merge everything in #38

PR #35, opened 2015-11-18
add a basic example

jlongster:
Let's start adding some example. Not going to merge this quite yet but just starting.


jlongster:
I just merged this under a different commit.

Issue #12, opened 2015-11-09
compatible with HashHistory?


box-turtle:
I see a strange behavior if I try to use this with HashHistory and call syncReduxAndRouter.

The queryKey param (http://rackt.org/history/stable/HashHistoryCaveats.html) seems to change every second or so, changing in the URL bar and I think it's reloading the route every second as well.

![image](https://cloud.githubusercontent.com/assets/3999743/11042322/60cf6c34-86e4-11e5-92cc-4580eb649c88.png)

I get this warning in console: 

Warning: You cannot PUSH the same path using hash history

It's possible this is something with my app, but I didn't see it when using redux-router.

I like this API much better! Just curious if anyone else has seen this behavior.

kadmil:
I've catch this too.

kadmil:
Looks like side-effect of putting location-with-hash back to history. That should be it: then you update history of createHashHistory it tries to save `pushState` to hash and trigger updates.

kadmil:
![image](https://cloud.githubusercontent.com/assets/4445523/11121071/e039b6be-8964-11e5-82f0-45f5917c447d.png)
We've forgot about `key` part

jlongster:
Thanks all, I'll try to look at this more today. Admittedly I don't understand the `key` part of HashHistory.

kadmil:
It's something they put as query param to save previous state, and it's quite puzzling.

zeroshine:
I have the same problem too.
I find it's the problem of the function locationToString
In createHashHistory routing path is /xxxx, not include /#
but in locationToString will return location.hash include /#
so it can't avoid history pushstate

jlongster:
There's a PR for this: #23. I'd like to find a prettier way to fix, but looks like we basically just need to do something different in `locationToString` if using hashHistory. Will look at it more soon.

amccloud:
@jlongster Does that PR disregard all query params from being stored in state?

Looks like the key param `_k` never makes it into the store so every time the store changes the syncing thinks the location changed because the key param is present in `window.location`. We could strip the key param from `window.location` before comparing to what is in the store. The key param can be renamed from `_k` so this would likely need to be a customizable configuration to handle it correctly.

I wish they would just remove state from history :disappointed:

zeroshine:
@amccloud 
In PR#23, I just write a simple regex in ~locationToString~ function to match the string wihout ~~?~~.
You can check its diff for more information
[...]

PR #40, opened 2015-11-22
Experimental sync logic

jlongster:
@mariocasciaro @kadmil @box-turtle @amccloud Looping you in because you were involved in #12 and #30.

This is an experiment to simplify things even further. So far we've had trouble being compatible with HashHistory and also triggering multiple updates of the same URL when transitioning. This should fix all of that.

This is also fixes an issue where multiple calls to `updatePath` *should* trigger multiple route updates. Right now if you click the same `<Link>` multiple times it does trigger a route update each time, even though it's going to the same URL. Previous behavior would not do this, but with this PR it does.

Can everyone try out this PR and let me know of any obvious problems? I've tested it in simple cases so far.


mariocasciaro:
I tried your experimental branch in my project, so far so good, it seems to be working fine.

ZucchiniZe:
Just tried this branch and it looks like it is working

ZucchiniZe:
Also, you probably want to change the main property in the package.json from `lib/index` to `src/index`

amccloud:
@ZucchiniZe lib/index is correct. Since you installed from github you need to run `npm run build`

ZucchiniZe:
Ah, that makes sense, thank you!

amccloud:
@jlongster works for me :+1:

jlongster:
Awesome! I think I'm going to land a simple test suite along with this so we can start adding tests over time as well. Still want to make sure this doesn't break anything.

heydemo:
@amccloud thanks for the heads up - Using this branch and works perfectly

n1k0:
Just confirming this patch makes the lib usable for me :)

jlongster:
I just added some tests. Let's go ahead and land this and see how it goes.

MrEfrem:
:+1:

PR #41, opened 2015-11-23
Support Immutable.js state

nuc:
Hey there,

I've added a small unwrapping of the state object, in case it is an Immutable.js instance.

Probably it's a bit too implementation specific, but it might be useful for others.

Cheers,
Giorgos

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,7 +38,13 @@ function locationToString(location) {
 
 function syncReduxAndRouter(history, store, selectRouterState = SELECT_STATE) {
   let lastRoute;
-  const getRouterState = () => selectRouterState(store.getState());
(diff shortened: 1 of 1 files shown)


jlongster:
We don't want to embed Immutable.js-specific logic in here. You can do this by passing a custom `selectRouterState` function:

```js
syncReduxAndRouter(history, store, state => state.get('routing'))
```

Thanks for the PR though, keep them coming!

jlongster:
(or maybe `state => state.get('routing').toJS()`)

jlongster:
(actually no, the router state will always be a plain JS object but if you are using immutable.js it'll be part of a Map, the first code should work)

PR #42, opened 2015-11-24
Update .npmignore

DLMR:
Made it so that the `.babelrc` file is not included when publishing to npm.

I have an application that is still on babel 5 and when building it finds the `.babelrc` inside `redux-simple-router` and freaks out since it does not understand `presets`.

This problem was introduced in `0.0.9`.

```
ERROR in ./~/redux-simple-router/lib/index.js
Module build failed: ReferenceError: [BABEL] .../node_modules/redux-simple-router/lib/index.js: Unknown option: .../node_modules/redux-simple-router/.babelrc.presets
```


jlongster:
Oh, thanks! This is adding a file that seems to ignore all files? Why the `.*` pattern?

jlongster:
I might prefer if you ignore just .babelrc and add a comment as to why. I'm assuming you can add comments with `#`?

DLMR:
I copied it from another project that I have, the `.*` pattern ignores all dot files. 

I will change it to explicitly ignore `.babelrc` along with a comment.

jlongster:
Thanks a lot! Will publish a patch release now.

PR #47, opened 2015-11-25
fix usage of updatePath

sokra:
`updatePath` takes a (true) `bool` as second param.

see signature of `function updatePath(path, avoidRouterUpdate)`


kjbekkelund:
This should be fixed when https://github.com/jlongster/redux-simple-router/pull/38 is merged

jlongster:
I believe #38 fixed it which I just merged. Sorry for the duplicate work @sokra and thanks a lot for the PR! That patch was a big one and I usually like to do smaller ones but that implemented a large feature.

sokra:
No problem

Issue #50, opened 2015-11-26
Stopping cycles


kjbekkelund:
There are two flows to manipulate the browser history:

```
1. Update URL/history
   -> history callback
   -> dispatch push/replace action
   -> store callback
```

```
2. Dispatch push/replace action
   -> store callback
   -> update history
   -> history callback
```

In both cases we call both of our callbacks, and in both cases we need to make sure that the flow stops before we create a cycle. E.g. let's say we want to change the URL by dispatching an action. This action triggers a history update, and when handling this update we have to make sure we don't dispatch yet another time.

To solve both cases without relying on `deep-equal` I think we need to track two variables — one to stop each flow.

To handle case 1 we set `{ avoidRouterUpdate: true }` in the history callback. As we don't increment the `changeId` we can stop the flow in the store callback by comparing it against the `lastChangeId`. By using `changeId` we also enable actions in `history.listenBefore` callbacks.

To handle case 2 we need to stop after we have updated the history based on a `pushState` or `replaceState` action. The problem is how to keep track of this update. In case 1 we decide to not increment the change id, thereby stopping the flow, but when updating the history I'm not sure how we can do the same. Right now we rely on comparing the current router state location with the new history location, but it would be nice to not need the `deep-equal` dependency in #38.

One option is to keep some kind of state in `location.state` in the same way as we have `changeId` in the store state, but I'm not sure if that would work (We don't have the same control over the `history` methods as we have over the `pushPath` and `replacePath` actions). Another option is to add another variable, e.g. `isRoutingFromAction`, which keeps track of whether or not the current route change was triggered by an action. Basically something like this:

```
let isRoutingFromAction = false;

const unsubscribeHistory = history.listen(location => {
  if (isRoutingFromAction) {
    // By handling this we have stopped the cycle, and
    // therefore need to reset the state.
    isRoutingFromAction = false;
    return;
  }

  // ...
});

const unsubscribeStore = store.subscribe(() => {
  const routing = getRouterState();
  if(lastChangeId === routing.changeId) return;

  isRoutingFromAction = true;

  // ...
});
```

With this change all tests still pass and my app still works, but as we don't run the tests in browsers I'm not entirely sure about potential edge cases (e.g. I don't use `listenBefore` at all in my app).

(I can create a PR with this if anyone wants to play around with it)

Maybe someone has other ideas about how to handle this?

kjbekkelund:
[...]

PR #52, opened 2015-11-27
Simple Karma config for browser testing

ellbee:
This is a work in progress to see if this is the direction we want to take for browser testing. You can specify the browsers you want to test with. By default it just runs the tests in Firefox.


karma.conf.js
```diff
@@ -0,0 +1,48 @@
+var path = require('path');
+var webpack = require('webpack');
+
+module.exports = function (config) {
+
+  // To specify browsers to test against use BROWSER var and comma separated list:
+  // BROWSER=Firefox,Chrome,Safari,IE npm run test:browser
+  var browsers = process.env.BROWSER ? process.env.BROWSER.split(',') : [ 'Firefox' ];
```
kjbekkelund: You can already specify browsers through the cli, e.g.

```
--browsers Chrome,Firefox
```

This overrides the default in the config file.

karma.conf.js
```diff
@@ -0,0 +1,48 @@
...
+      entry: path.join(__dirname, 'test', 'index.js'),
+      module: {
+        preLoaders: [
+          {
+            test: /\.js$/,
+            exclude: /node_modules/,
+            loader: 'babel',
+            query: {
+              presets: ['es2015'],
+              plugins: ['transform-object-assign']
```
kjbekkelund: Is it possible to rely on both of these specified in `.babelrc`? Less copy-pasting and things that must be in sync across files.

karma.conf.js
kjbekkelund: Hm, haven't tested it with npm scripts though. Does `npm run test:browser --browsers Chrome,Firefox` do the right thing?

karma.conf.js
kjbekkelund: (I guess that also means that `--plugins transform-object-assign` should be removed from https://github.com/jlongster/redux-simple-router/blob/0941951a0a6948e526b33d49f49dfda033cbe6f0/package.json#L15, so maybe it should be its own PR?)

kjbekkelund:
Great stuff, @ellbee. I'll try to test this with #38 and #50.

karma.conf.js
kjbekkelund: Played around with it, using `npm run test:browser -- --browsers Chrome,Firefox` works

ellbee: No, you can't do `npm run test:browser --browsers Chrome,Firefox`.

However, it seems that since npm 2 you **can** do `npm run test:browser -- --browsers Chrome,Firefox` which I had no idea had been added until 5 mins ago. Great!

karma.conf.js
ellbee: Yes, good point.

ellbee:
Thanks for the review @kjbekkelund!

Usage has changed to:

`npm run test:browser` to run tests in firefox
`npm run test:browser -- --browsers Chrome,Firefox` to specify browsers

@jlongster The `transform-object-assign` plugin is now specified in the `.babelrc` and removed from the command line args of `npm run build`. @kjbekkelund suggested that this maybe should be a separate pull request, let me know if you want me to separate it out.

kjbekkelund:
If I change `createMemoryHistory` to `createHashHistory` 5 tests fail. I'm on a flight for 10 hours tomorrow, so I'll try to take a look at it. Would be awesome to have tests with `createHashHistory` work as expected.

kjbekkelund:
Ah, it was actually quite easy to get the tests running:

```diff
-const { createMemoryHistory: createHistory } = require('history');
+const { createHashHistory, createMemoryHistory } = require('history');
+
+let isBrowser = false;
+let createHistory = createMemoryHistory;
+if (typeof window !== 'undefined') {
+  isBrowser = true;
+  createHistory = createHashHistory;
+}
+
+beforeEach(() => {
+  if (isBrowser) {
+    window.location = '#/';
+  }
+});
```

It looks a little hacky, so we might want to clean it up, but at least everything is green. Maybe we should at it to this PR? What do you think, @jlongster? I can of course create a PR when this is merged if you want that.

ellbee:
@kjbekkelund Nice, thanks for figuring out the failures. Turns out browser history was failing too. All tests are passing now. 

I have set it up so that test:node runs the tests with memory history and test:browser runs the tests with both browser history and hash history. If there are any tests we only want to run in node or the browser then they can be put in 'test/node' or 'test/browser' and will be picked up automatically.

ellbee:
Rebased onto master.

jlongster:
Sorry for the delay. Work has been *especially* crazy. (if it's any consolation, I just landed a big firefox debugger patch and it officially uses redux quite extensively now!)

I hadn't looked at this and I didn't realize you moved all the tests, and I just merged another PR that updated the actions to the FSA-compliant :( It's not that hard to change, all the tests just need to check `action.payload` instead of the `action` itself. Do you mind rebasing this PR or do you want me to do it?

ellbee:
Hey, no worries! I don't mind rebasing, I'll get it done right now.

Not that I needed a consolation, but Firefox debugger using redux is really awesome! :+1:

jlongster:
Wow, thanks! Looks great to me, I've never used karma before so good to learn. I'll merge this now if you think it's ready?

ellbee:
Yeah, I think it is ready. Although I have used it in the past, I am by no means a Karma expert myself. I stole pretty liberally from the test setup in React Router and History so I am confident that this is along the right lines.

jlongster:
Awesome, I'll merge and then try it out.

There's another PR suggesting setting up TravisCI, I wonder if I can run browser stuff there.

ellbee:
I think Travis has Firefox, so should be fine.

PR #55, opened 2015-11-28
Use ES6 default params for opts

kjbekkelund:


MattKunze:
You can add ES6 destructuring and make the signature entirely self-documenting as well:

`function replacePath(path, state, {avoidRouterUpdate} = {})`

PR #56, opened 2015-11-28
Ensure correct push/replace state in store

kjbekkelund:
I think we should be consistent about having `replace` correct in the store. It doesn't change much, but it _might_ be relevant if you play around with `UPDATE_PATH` in a reducer.

As far as I can tell `history` treats same url but different state as a replace:

```
history.pushState(null, '/foo');
// to
history.pushState({ bar: 'baz' }, '/foo');
```

I'm also open to just skip this one if we don't want the added "complexity".

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -116,28 +116,35 @@ describe('syncReduxAndRouter', () => {
     history.pushState(null, '/foo');
     expect(store.getState().routing.path).toEqual('/foo');
     expect(store.getState().routing.state).toBe(null);
+    expect(store.getState().routing.replace).toBe(false);
 
     history.pushState({ bar: 'baz' }, '/foo');
     expect(store.getState().routing.path).toEqual('/foo');
     expect(store.getState().routing.state).toEqual({ bar: 'baz' });
+    expect(store.getState().routing.replace).toBe(true);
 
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -116,28 +116,35 @@ describe('syncReduxAndRouter', () => {
...
     expect(store.getState().routing.path).toEqual('/bar?query=1');
+    expect(store.getState().routing.replace).toBe(false);
 
     history.replaceState({ bar: 'baz' }, '/bar?query=1');
     expect(store.getState().routing.path).toEqual('/bar?query=1');
     expect(store.getState().routing.state).toEqual({ bar: 'baz' });
+    expect(store.getState().routing.replace).toBe(true);
 
-    history.pushState(null, '/bar?query=1#hash=2');
+    history.pushState({ bar: 'baz' }, '/bar?query=1#hash=2');
```
kjbekkelund: I kept the same state as the previous state change so _only_ the url changed in the `pushState`.

jlongster:
Definitely agree. This looks good, thanks!

jlongster:
I'm going to be going through this code and doing more testing this week, and we can release it by the end of the week hopefully.

PR #63, opened 2015-11-30
use FSA actions

nishp1:
PR for https://github.com/jlongster/redux-simple-router/issues/53.

This is a breaking change for folks who have a reducer for constant `UPDATE_PATH`. Bump version?


jlongster:
Yeah, we're already breaking the API in the next version. Might as well call it 1.0.0. Thanks for this.

PR #74, opened 2015-12-04
Restore path on Devtools reset

ellbee:
I created a pull request here for discussion on how to fix [this problem](https://github.com/jlongster/redux-simple-router/pull/38#issuecomment-159653786) instead of cluttering up #73. This is my current best effort. I tried the approach of setting initialState to be null (or false etc.) initially and then filling it in when history.listen callback fired, but it caused a bunch of errors later on in other functions that were expecting the normal keys.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,14 +33,20 @@ function replacePath(path, state, { avoidRouterUpdate = false } = {}) {
 
 // Reducer
 
-const initialState = {
-  changeId: 1,
-  path: undefined,
-  state: undefined,
-  replace: false
-};
+// The initial path is needed when resetting the state in Devtools but is
+// unknown until the history.listen callback set up in syncReduxAndRouter fires
+let initialRoute;
 
-function update(state=initialState, { type, payload }) {
+function initialState() {
+  return {
+    changeId: 1,
+    state: undefined,
+    replace: false,
+    path: initialRoute && initialRoute.path
+  }
+}
+
+function update(state=initialState(), { type, payload }) {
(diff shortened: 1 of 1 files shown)


jlongster:
Why not just overwrite `initialState` entirely? I think this makes it more confusing than just resetting the whole thing.

ellbee:
Yeah, probably. The idea was to give an indication that something slightly weird is going on with the path in the `initialState`, rather than just overwriting it in a different part of the code.

I'll close this out. The solution for this was merged has a slight problem, but I'll open an issue for that.

jlongster:
In general though I like changing (not mutating, just resetting a var) the initialState more than how we currently do it. Currently in the store subscriber we check if we are using the initialState and it so, grab the first route loaded from `history.listen`. This causes inconsistent state at the beginning though; the state in the app state will not reflect the current route.

jlongster:
I actually like how this is headed though more than how we currently do it. I'd rather overwrite `initialState` instead of magically using a *different* state when `initialState` is passed in `store.subscribe`.

ellbee:
Is this along the lines of what you are thinking? https://github.com/jlongster/redux-simple-router/pull/73#issuecomment-161938264

jlongster:
Close, but I'd rather not depend on state in `initialState` to see if we should reset it or not. I might just make another local variable to track if it's the first load.

ellbee:
Ah, I see. That sounds good to me.

jlongster:
See this now: https://github.com/jlongster/redux-simple-router/pull/83

PR #77, opened 2015-12-05
Create test helper to improve tests

kjbekkelund:
I think this nicely cleaned up a couple of the tests. Plus it's easier to not specify the `changeId` while still keeping the tests easily readable.

### DIFF
--- a/test/createTests.js
+++ b/test/createTests.js
@@ -2,6 +2,25 @@ const expect = require('expect');
 const { pushPath, replacePath, UPDATE_PATH, routeReducer, syncReduxAndRouter } = require('../src/index');
 const { createStore, combineReducers } = require('redux');
 
+expect.extend({
+  toContainRoute({
+    path,
+    state = undefined,
+    replace = false,
+    changeId = undefined
+  }) {
+    const routing = this.actual.getState().routing;
+
+    expect(routing.path).toEqual(path);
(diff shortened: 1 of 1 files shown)


ellbee:
Very nice! :+1:

test/createTests.js
```diff
@@ -145,127 +164,134 @@ module.exports = function createTests(createHistory, name, reset = defaultReset)
...
+        expect(store).toContainRoute({
           path: '/foo',
-          changeId: 2,
           replace: false,
           state: undefined
         });
 
         store.dispatch(pushPath('/foo'));
-        expect(store.getState().routing).toEqual({
+        expect(store).toContainRoute({
```
jlongster: I think I like this. `changeId` is a case where it might be too much implementation detail to encode in tests.

This is one place where we lose information though. You aren't testing that `history.pushState` is actually called twice.

kjbekkelund: Ah, good catch, @jlongster. I'll fix asap.

kjbekkelund:
@jlongster This should cover it, I think.

jlongster:
Cool. (I was wrong in that the place that actually tested it was later in the file)

jlongster:
I promise to look at the other PRs and resolve the devtools issues today or tomorrow!

PR #81, opened 2015-12-06
Don't navigate immediately on boot.

STRML:
Had an issue in my app, in combination with `react-fetcher`, where `redux-simple-router` would create a history navigation on load, even though there's no reason to immediately navigate right on boot (the url is already in the URL bar).

Was this behavior intentional? Changing this line fixed the issue for me and I no longer have a redundant `UPDATE_PATH` on boot and an additional fetch.

If it helps, I'm pairing this with populating `state.routing.path` on the server side using `req.url`, so the routing path is correctly pre-populated on boot.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -114,7 +114,7 @@ function syncReduxAndRouter(history, store, selectRouterState = SELECT_STATE) {
 
     // Only trigger history update is this is a new change or the location
     // has changed.
-    if(lastRoute !== undefined &&
+    if(lastRoute === undefined ||
       lastRoute.changeId === routing.changeId &&
       locationsAreEqual(lastRoute, routing)) {
       return;


jlongster:
The problem is that there's no way to get the "current location", so `initialState` has the wrong URL (it just defaults to `/`). It will dispatch one action to update the store with the current URL.

With your changes here, the app state will not have the correct route for the initial route.

jlongster:
We can get around this by creating a different action type, I suppose, something like INIT_PATH.

STRML:
Actually, it seems that it works fine. I'm not preventing the state from updating (`UPDATE_PATH` still fires), it just prevents a useless `history.pushState()`.

For example (PENDING, FULFILLED actions are from `react-fetcher` on the route component):

screen shot 2015-12-06 at 12 45 18 pm

jlongster:
Ah, I see what you mean. That makes sense. I'm actually making some changing in this area but I'll go ahead and merge this.

STRML:
Thanks!

jlongster:
I think this actually broke some tests. I need to integrate TravisCI on here.

I'll look into it.

jlongster:
Ok, so I need to back this out because it broke tests, most importantly these 2:

https://github.com/jlongster/redux-simple-router/blob/master/test/createTests.js#L436
https://github.com/jlongster/redux-simple-router/blob/master/test/createTests.js#L505

Please wait a day or two because I'm going to land some changes with this code, and it might fix your issue.

STRML:
No worries. Thanks.

jlongster:
See #83 now.

PR #82, opened 2015-12-06
update comments and avoid early returns

jlongster:
I added some info to the comments because I wanted to explain every detail as to how things work. I also change the code to remove early returns, as I think it is clearer.

Just opening a PR in case I made any mistakes, but I'd like to merge soon. All tests still pass.


jlongster:
Going to merge this because there's another important PR I want to make after it

PR #89, opened 2015-12-07
Update History and DevTools

### LINKED ISSUE #69: Redux-simple-router and React-Router replaceState interactions
Hello! I need to implement authentication mechanics to my application and so instead of starting from scratch I've used [React-Router auth-flow example](https://github.com/rackt/react-router/tree/master/examples/auth-flow) as example.

I'm facing a strange problem: from the `Login` component (roughly the same as the example above) I should be able to redirect to the previous visited route if auth is successfull:

```javascript
if (loggedIn) {
   const { location, history } = this.props;

   if (location.state && location.state.nextPathname) {
       history.replaceState(null, location.state.nextPathname);
   } else {
       history.replaceState(null, '/');
   }
}
``` 

This path should be stored inside the location state object by the `onEnter` hook on a protected route:

```javascript
function requireAuth(nextState, replaceState) {
    if (! isLoggedIn()) {
        replaceState({ nextPathname: nextState.location.pathname }, '/login');
    }
}
```

If I use this logic with `redux-simple-route`, inside the `Login` component the props `this.location.state` is `null`, while, if disabled, contains the right nextPathname property.

I think I've narrowed down the problem to these lines of `redux-simple-router` (68-71): 

```javascript
if(lastChangeId !== routing.changeId) {
  lastChangeId = routing.changeId;
  history.pushState(null, routing.path);
}
```

It pushes the route to history (which it should be not necessary) and doesn't pass the state in the process.

A simple fix would be setting `changeId` in the initial state to `0` but I don't know if it may cause other things to stop working...

justingreenberg:
Resolves #69 #92 

The example is updated and routing/devtools are working as expected—routing remains synced with store and devtools, instrumentation toggles sync with routing (reset, revert, sweep), etc

A couple of things to consider:
 * Should the state keys be updated match `react-router` (ie, `pathname` instead of `path`)
 * Should we include including [query support](https://github.com/rackt/history/blob/master/docs/QuerySupport.md) now that it's part of the `LocationDescriptorObject`? i know it's a common feature request and would be trivial to implement

Based on the new API, I'm thinking we should just mirror the whole **[`history.LocationDescriptorObject`](https://github.com/rackt/history/blob/master/docs/Glossary.md#locationdescriptor)**—it's very simple and contains all relevant state in `window.location`...
```js
state.routing = { pathname, search, query, state }; // mirror history.LocationDescriptorObject 
```
@jlongster @kjbekkelund ellbee what do you guys think?


src/index.js
```diff
@@ -136,8 +134,12 @@ function syncReduxAndRouter(history, store, selectRouterState = SELECT_STATE) {
        !locationsAreEqual(lastRoute, routing)) {
 
       lastRoute = routing;
-      const method = routing.replace ? 'replaceState' : 'pushState';
-      history[method](routing.state, routing.path);
+
+      const method = routing.replace ? 'replace' : 'push';
+      const locationParam = (!!routing.state && typeof routing.path === 'String') ? 
+        routing.path : { state: routing.state, pathname: routing.path}
```
kjbekkelund: Is there a reason for switching between string and object based on `state`? Much easier to read the code if we can always send an object.

test/createTests.js
```diff
@@ -1,8 +1,8 @@
 const expect = require('expect');
 const { pushPath, replacePath, UPDATE_PATH, routeReducer, syncReduxAndRouter } = require('../src/index');
 const { createStore, combineReducers, compose } = require('redux');
-const { devTools } = require('redux-devtools');
-const { ActionCreators } = require('redux-devtools/lib/devTools');
+const React = require('react');
```
kjbekkelund: Is `React` needed in the tests? Don't see anything related to it in the diff.

src/index.js
justingreenberg: I agree we can just send the object, it works both ways

src/index.js
```diff
@@ -53,16 +53,14 @@ let initialState = {
...
-      state: payload.state,
-      replace: payload.replace
-    });
+    const { path, state, replace, avoidRouterUpdate } = payload;
+    const changeId = newState.changeId + (avoidRouterUpdate ? 0 : 1);
+
+    return Object.assign({}, newState, { path, state, replace, changeId });
   }
-  return state;
+  return newState;
```
kjbekkelund: I actually think this is more difficult to read with all the destructuring.

kjbekkelund:
I like the idea of renaming `path` to `pathname` to be in sync with `history`

src/index.js
```diff
@@ -92,7 +90,7 @@ function syncReduxAndRouter(history, store, selectRouterState = SELECT_STATE) {
 
   const unsubscribeHistory = history.listen(location => {
     const route = {
-      path: history.createPath(location),
+      path: history.createHref(location),
```
kjbekkelund: Why is this needed?

.gitignore
```diff
@@ -1,3 +1,5 @@
 lib
 node_modules
 coverage
+npm-debug.log
+examples/basic/npm-debug/.log
```
kjbekkelund: I think there's a typo in the last one? And I don't think it's needed when you already have `npm-debug.log` there.

src/index.js
justingreenberg: it is needed to extract the url encoded path for current implementation. if we decided to mirror `history` location objects with query params, this line will not be necessary

justingreenberg:
@kjbekkelund thanks for reviewing so quicky. it's like 5am in nyc i'm gonna crash. will pick this up tomorrow

src/index.js
ellbee: This is the line that is breaking the tests in the browser. Changing back to `createPath` makes the tests pass.

test/browser/index.js
```diff
@@ -1,5 +1,5 @@
-const { createHashHistory, createHistory } = require('history');
+const { createHashHistory, createBrowserHistory } = require('history');
 const createTests = require('../createTests.js');
 
 createTests(createHashHistory, 'Hash History', () => window.location = '#/');
-createTests(createHistory, 'Browser History', () => window.history.replaceState(null, null, '/'));
+createTests(createBrowserHistory, 'Browser History', () => window.history.pushState(null, null, '/'));
```
ellbee: Changing `createHistory` to `createBrowserHistory` breaks the tests. History does not export `createBrowserHistory` so it is `undefined`.

ellbee:
Do we want to be using beta releases in the examples? We should bear in mind that people are going to be treating this code as best practice. I will add that I've been using the Devtools beta release and think that it is solid. Maybe the decision to use a beta should be on a case by case basis.

This PR also bumps webpack to the 2.0 beta release, can we justify this? It gives off some peer dependency warnings on `npm install`

kjbekkelund:
@ellbee I agree — I thought about raising the same issue. I'm generally not a big fan of relying on beta releases in code people will copy when learning something new. I would prefer this PR without the beta releases, especially the Webpack upgrade.

justingreenberg:
i agree that `redux-devtools@3` is solid. in terms of webpack, i don't have much experience with v2, but i do know it's used in production and obviously has a solid team behind it

there's almost no change in api surface so very easy to revert—it's totally up to you guys

src/index.js
justingreenberg: i tried `createPath` but devtools tests were still failing... did you update deps?

ellbee: Whoops, missed something... See new comment.

test/createTests.js
```diff
@@ -160,14 +159,15 @@ module.exports = function createTests(createHistory, name, reset = defaultReset)
 
       beforeEach(() => {
         history = createHistory();
-        const finalCreateStore = compose(devTools())(createStore);
+
+        const finalCreateStore = compose(instrument())(createStore);
         store = finalCreateStore(combineReducers({
           routing: routeReducer
         }));
         devToolsStore = store.devToolsStore;
```
ellbee: Should be `store.liftedStore`

kjbekkelund: :+1:

gaboesquivel:
:+1:

justingreenberg:
`routing.path` :point_right: `routing.pathname` to match `history.location` object keys

@jlongster @ellbee if we're syncing pathname and state, why not sync the full [LocationDescriptor](https://github.com/rackt/history/blob/master/docs/Glossary.md#locationdescriptor)?

```ts
type LocationDescriptorObject = {
  pathname: Pathname;
  search: Search;
  query: Query;
  state: LocationState;
};

type LocationDescriptor = LocationDescriptorObject | Path;
```

it's super minimal, and there's really nothing in there that would pollute the redux store. [query support](https://github.com/rackt/history/blob/master/docs/QuerySupport.md) could easily be optional but supported... users could simply use the [`useQueries`](https://github.com/rackt/history/blob/master/docs/Glossary.md#createhistoryenhancer) HOC:

```js
import { createHistory, useQueries } from 'history'
const history = useQueries(createHistory)()
```

`history.useQueries` is an enhancer that passes parsed query strings to `history.location` (LocationDescriptor above) and stringifies outgoing queries when creating or pushing paths:

```js
history.createPath({ pathname: '/the/path', query: { the: 'query' } })
history.push({ pathname: '/the/path', query: { the: 'query' } })
```

please let me know what you guys think

kjbekkelund:
It definitely seems easier to support query params with the new release of `history`. If we do that change we should probably change `pushPath` and `replacePath` to parameters on the same format as `history.push`/`replace`, otherwise we need to keep adding stuff to the api.

It ends up being a huge change. I would rather see a clean PR with `push`+`replace`, plus the needed updates to get that working, but without all the other changes, then new PRs with the other stuff (upgrade to betas of devtools and webpack, change to `pathname`, support for queries, etc). That makes it far easier to be sure that we've thought through everything and added the necessary new tests etc.

ellbee:
I agree with @kjbekkelund on both points.

justingreenberg:
ok cool i moved README updates to #93, removed webpack beta and reverted `path` update

test/browser/index.js
```diff
@@ -2,4 +2,4 @@ const { createHashHistory, createHistory } = require('history');
 const createTests = require('../createTests.js');
 
 createTests(createHashHistory, 'Hash History', () => window.location = '#/');
-createTests(createHistory, 'Browser History', () => window.history.replaceState(null, null, '/'));
```
kjbekkelund: Any reason this is needed?

kjbekkelund:
Nice, I'll test it out later today :)

jlongster:
I don't have time to go through all of this right now, but I'm a little worried because right now we don't tie ourselves to any version of `history`. I thought the `pushState` and such methods were a stable API; now that those are changing we need to actually use peer dependencies or something. We can't just assume any version of redux-simple-router will work with any version of history.

Which kind of sucks. It gets tricky to make sure the user is using the right version of history with the right version of this lib. We need to look into how to make that work.

justingreenberg:
@jlongster no worries! this pr just brings things up to date with `react-router@1.0.2` and the new `history` method names (stops the constant warnings for createHref, pushState, replaceState, etc.)

>Now that we’ve hit 1.0, we are dedicated to smooth upgrade paths — @ryanflorence

since rackt [is implementing more formal versioning](https://medium.com/rackt-and-roll/upgrading-and-backwards-compatibility-for-react-router-and-other-rackt-repos-ec24978a58b2#.p7c866ny7) it probably makes sense for `redux-simple-router` to track it's siblings (and `history@2.0.0` will be shipping soon)

in terms of adding new features, ie. support for query and search, @kjbekkelund started a conversation and outlined some implementation options in #95 :smile:

kjbekkelund:
@justingreenberg As we need to think more about how to handle the `history` upgrade, maybe you could split this into two PRs? One for devtools and one for history, then we could look at getting the devtools stuff in now at least.

justingreenberg:
@kjbekkelund no problem, i have to step away from my desk for a hour or so but i'll move devtools updates into new PR when i get back :)

i still think it makes sense to merge these history updates asap so RSR can continue to be used as-is without deprecation notices (with minor version bump) until there's a decision on history upgrade

justingreenberg:
@kjbekkelund now that devtools is stable i'll submit new PR in the morning for devtools update

@jlongster what are you thinking in terms of other updates to sync up with new history api? since there are merge conflicts and i'm splitting out devtools anyway, would you like me to to close this PR, clean them up and submit new pr? please let me know :)

kjbekkelund:
@justingreenberg Awesome :100:

jlongster:
@justingreenberg You can create a new clean PR with just updating the history integration. Just make sure to link this one so it's easy to follow the discussion.

I have more time now so when you open a new PR I'll take a closer look at it.

jlongster:
Also I agree that we should update to whatever the newest `history` API is internally and we can open a separate issue for possible changes to our API to better support location objects if we want to do that.

PR #96, opened 2015-12-09
Exporting INIT_PATH constant

tonovotny:
So we can use this action type in reducers

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -149,6 +149,7 @@ function syncReduxAndRouter(history, store, selectRouterState = SELECT_STATE) {
 }
 
 module.exports = {
+  INIT_PATH,
   UPDATE_PATH,
   pushPath,
   replacePath,


kjbekkelund:
Do you have a use case for this? It wasn't intended to be used in reducers.

tonovotny:
I'm using it to modify state according to action.payload.path value. Maybe the action.payload is not intended to be used externally (?)

kjbekkelund:
You should use `UPDATE_PATH` for that. The `INIT_PATH` is only intended to be used internally to set up the internal state we need for Redux DevTools to work as expected.

tonovotny:
Yes, I do. The problem is getting the initial value and my idea was simply using `INIT_PATH` action. Of course I can get it elsewhere but it's not so nice then..

kjbekkelund:
Good point, @tonovotny. Actually, looking at the code now I'm not entirely sure we need `INIT_PATH` at all anymore... I created #97, which removes it. Let's see what @jlongster says.

jlongster:
I'm at a work week right now and things are crazy busy :( I think I can find some time today to take a look though. No comments yet!

kjbekkelund:
@jlongster If you've got some time, focus on https://github.com/jlongster/redux-simple-router/pull/93. That's the only thing left for a 1.0. The rest we can get into later versions.

jlongster:
I actually think it's fine to listen for `INIT_PATH` if you are wanting to set up some state that depends on the current URL. But I'm not sure. It may be best just to always emit `UPDATE_PATH`.

kjbekkelund:
I'm in favour of `UPDATE_PATH`. But it's a breaking change, so open for both solutions. It shouldn't be a problem to "imitate" `INIT_PATH` in a reducer (only do something the first time you receive an `UPDATE_PATH`), so I prefer the "smaller API surface".

tonovotny:
I like small API surface too but would prefer unambiguous solutions, so `INIT_PATH`. But I'm OK with always emiting `UPDATE_PATH`.

kjbekkelund:
Any examples of use cases where init is important, instead of just listening for update?

kjbekkelund:
If we compare [these two tests](https://github.com/jlongster/redux-simple-router/blob/0145d6f3599efd4a66db33f6ae8b596b9aa455dd/test/createTests.js#L492-L518) I find it natural for `UPDATE_PATH` to work the same way. When you `history.listen` you first get the current url, then you get updates from this url.

jlongster:
Closing in favor of #97

PR #97, opened 2015-12-09
Remove `INIT_PATH`

kjbekkelund:
No tests fail, so not sure we need this constant? We should either add a failing test or remove it.

The use case for removing this is receiving the initial URL when using `UPDATE_PATH`, e.g. #96.

(So the functional difference is: do we notify the initial route or only changes from the initial route?)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,23 +2,8 @@ import deepEqual from 'deep-equal'
 
 // Constants
 
-const INIT_PATH = '@@router/INIT_PATH'
 export const UPDATE_PATH = '@@router/UPDATE_PATH'
 const SELECT_STATE = state => state.routing
-
-// Action creators
-
-function initPath(path, state) {
-  return {
-    type: INIT_PATH,
-    payload: {
-      path: path,
-      state: state,
-      replace: false,
-      avoidRouterUpdate: true
-    }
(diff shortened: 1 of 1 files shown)


jlongster:
Thanks! Yes, it's just a functional difference. I did not add any tests for this functional difference yet. I'm not really sure which is more expected for the user; I'm fine removing it if we think `UPDATE_PATH` should be fired for the initial path too. (that should definitely be documented if so)

src/index.js
```diff
@@ -118,7 +105,8 @@ function syncReduxAndRouter(history, store, selectRouterState = SELECT_STATE) {
       // trigger an unnecessary `pushState` on load
       lastRoute = initialState;
 
-      store.dispatch(initPath(route.path, route.state));
+      const method = location.action === 'REPLACE' ? replacePath : pushPath;
+      store.dispatch(method(route.path, route.state, { avoidRouterUpdate: true }));
```
jlongster: Here we can just directly use `pushPath` because we are avoiding a route update anyway.

jlongster:
Can you rebase this? I will merge it in.

kjbekkelund:
@jlongster Rebased

Issue #103, opened 2015-12-10
1.0.0 breaks if history uses `basename` option


dtryon:
Hi,
We are currently using `react-router` version 1.0.0 `createHistory` which imports the browser history strategy.

With 0.0.10 all was fine, but upgrading causes a `pushState` to loop forever.  :(

We are wiring up `basename` like this:

```
const history = useBasename(createHistory)({basename: '/foobar'});
```

Now, when we use `pushState` it loops.  I think that the problem is in this code:

```
} else if(!locationsAreEqual(getRouterState(), route)) {
      // The above check avoids dispatching an action if the store is
      // already up-to-date
      const method = location.action === 'REPLACE' ? replacePath : pushPath
      store.dispatch(method(route.path, route.state, { avoidRouterUpdate: true }))
    }
```

The `!locationsAreEqual` always returns `true` because the call to `history.createPath(location)` comes back with the `basename` prepended.

If I debug in the browser, I can see it looping over this case.

I think the solution would be to add this type of thing `lastRoute.changeId !== routing.changeId` like in `unsubscribeStore`.

I'll see if I can reproduce in a unit test and issue a PR.

hunterc:
Running into the same issue with #102

kjbekkelund:
Interesting. I'm taking a look too, @hunterc @dtryon. Would be awesome if you're able to create a failing unit test for it!

dtryon:
We have a failing test:

```
it('handles basename history option', () => {
        const store = createStore(combineReducers({
          routing: routeReducer
        }))
        const history = useBasename(createHistory)({basename:'/foobar'});
        syncReduxAndRouter(history, store);

        store.dispatch(pushPath('/bar'))
        expect(store).toContainRoute({
          path: '/bar',
          changeId: 2,
          replace: false,
          state: undefined
        })
      })
```

kjbekkelund:
@dtryon Nice, thanks. I got @hunterc's example to run locally with this change:

```diff
diff --git a/src/index.js b/src/index.js
index 10cb737..495ec76 100644
--- a/src/index.js
+++ b/src/index.js
@@ -71,6 +71,16 @@ function locationsAreEqual(a, b) {
   return a != null && b != null && a.path === b.path && deepEqual(a.state, b.state)
 }
 
+function createPath(location) {
+  const { pathname, search, hash } = location
+  let result = pathname
+  if (search)
+    result += search
+  if (hash)
+    result += hash
+  return result
+}
+
[...]

Issue #95, opened 2015-12-09
Support query params?


kjbekkelund:
This morning I've been playing around with how we _could_ support query params in `redux-simple-router` (just as a thought experiment).

In `history.listen` we receive a `Location`. With `pushPath` and `replacePath` we can change the API to accept `LocationDescriptor`s. However, with no way of creating a `LocationDescriptor` object from a `LocationDescriptor` string, it's difficult to support query params.

Here's an example of the problem:

```
pushPath('/foo?a=b')
// store:
{ location: '/foo?a=b' }


pushPath({ pathname: '/foo', query: { a: 'b' } })
// store:
{ location: { pathname: '/foo', query: { a: 'b' } } }


history.push('/foo?a=b')
// store:
{ location: { pathname: '/foo', query: { a: 'b' } } }


<Link to='/foo?a=b'>
// store:
{ location: { pathname: '/foo', query: { a: 'b' } } }
```

The problem is of course that every time you `pushPath` you will get the exact `LocationDescriptor` you pass in into the store, but every time you `history.push` or `<Link>` you'll get an object. We then can't reliably support query params. (If you're very strict about always passing in a `LocationDescriptorObject` you would have query support, but that feels very fragile.)

We _can_ of course use `history.createLocation` right now, but it's most likely [no longer going to be public](https://github.com/rackt/history/pull/182) in `history` V2.

To support query params we need to be able to create a `LocationDescriptor` object both from a `LocationDescriptor` string and from a `Location`. We could then easily support query params, and we could also `deepEqual` check them without knowing about `query` or anything else that's actually part of the `LocationDescriptor` (which is how we stop cycles, #50). With a `createLocationDescriptor` (that always returns a `LocationDescriptorObject`) our code base would actually be simpler, and we would get query support for free.

Okey, so some potential solutions:

## Keep the solution we have now

I.e. `path` and `state` in store, but change to accepting `LocationDescriptor` in `push` and `replace`. We don't support query params.

## Verbatim `LocationDescriptor` in store

We have to make it _very_ clear that this is _either_ an object or a string. So it's most likely not very usable for people. We "support" query params, iff you're very strict about always passing in a `LocationDescriptorObject`. Fragile.

## Always go through `history` "before" store

```
pushPath
-> UPDATE_PATH_INTERNAL action
-> store.subscribe callback
-> history.push
-> history.listen callback
-> UPDATE_PATH action
-> store.subscribe callback where we break cycle
```

[...]

PR #104, opened 2015-12-10
Fix basename path issues

### LINKED ISSUE #102: IndexRedirect causes infinite route
If I use an `IndexRedirect` with a basename I seem to get an infinite route.
Filing it here because it only happens when I call `syncReduxAndRouter`.

Commenting out the basename seems to resolve the issue ( I assume because there is nothing to repeat).

If I hit `/test`, I get a infinite loop of `/test/test/test/......` until the call stack is exceeded.

[Gist I'm using to verify](https://gist.github.com/hunterc/7c8896f7ff05f1928eb5)

kjbekkelund:
Closes #102 #103 

The problem with relying on `history.createPath` is that it might change in ways we don't want it to, e.g. as here when `history.createPath` prepends the basename, which means that the input to `pushPath` will differ from the path we create on a location change.

I think a good first step is to create the path ourselves (as we did in the previous version), and going forward to think about the ideas in #95 for later versions.

Thanks @dtryon and @hunterc for helping to fix this

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,6 +71,16 @@ function locationsAreEqual(a, b) {
   return a != null && b != null && a.path === b.path && deepEqual(a.state, b.state)
 }
 
+function createPath(location) {
+  const { pathname, search, hash } = location
+  let result = pathname
+  if (search)
+    result += search
+  if (hash)
+    result += hash
+  return result
+}
+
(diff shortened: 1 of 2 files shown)


kjbekkelund:
@jlongster Should we get out a `1.0.1` that solves this?

PR #112, opened 2015-12-10
Rely less on changeId in tests

kjbekkelund:
Tried to clean up the tests so they have less duplication and rely less on checking `changeId`, but focus more on checking what actually triggers `history.listen` (i.e. the external api for these changes). This should make it easier to tests changes to how we stop cycles.

I removed the test named "does not unnecessarily update the store", but added a "only triggers once when updating path via store" (which I created as part of #109, where I have a test which covers `history.pushState`).

### DIFF
--- a/test/createTests.js
+++ b/test/createTests.js
@@ -357,39 +357,21 @@ module.exports = function createTests(createHistory, name, reset = defaultReset)
       })
 
       it('updates the router even if path is the same', () => {
-        expect(store).toContainRoute({
-          path: '/',
-          changeId: 1,
-          replace: false,
-          state: undefined
+        const updates = []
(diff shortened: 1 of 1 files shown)


jlongster:
> I removed the test named "does not unnecessarily update the store", but added a "only triggers once when updating path via store" (which I created as part of #109, where I have a test which covers history.pushState).

That doesn't seem to do the same thing as the test you removed. The test you removed listens for store changes, while the test you added listens for history changes.

But this looks good to me!

kjbekkelund:
Yeah, good point. I'll add another test for it.

kjbekkelund:
I added a test that covers store updates.

PR #131, opened 2015-12-19
chore(history): update to history ^0.14.0 internally

justingreenberg:
original discussion #89 i tried to keep changes as minimal as possible since it sounds like there may be a rewrite happening soon

i can also submit a separate PR for devtools now that it's stable if you want

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -88,7 +88,7 @@ export function syncReduxAndRouter(history, store, selectRouterState = SELECT_ST
   const unsubscribeHistory = history.listen(location => {
     const route = {
       path: createPath(location),
-      state: location.state === null ? undefined : location.state
+      state: location.state
     }
 
     if (!lastRoute) {
--- a/src/index.js
+++ b/src/index.js
@@ -88,7 +88,7 @@ export function syncReduxAndRouter(history, store, selectRouterState = SELECT_ST
   const unsubscribeHistory = history.listen(location => {
     const route = {
       path: createPath(location),
-      state: location.state
+      state: location.state === null ? undefined : location.state
     }
 
     if (!lastRoute) {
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -88,7 +88,7 @@ export function syncReduxAndRouter(history, store, selectRouterState = SELECT_ST
   const unsubscribeHistory = history.listen(location => {
     const route = {
       path: createPath(location),
-      state: location.state
+      state: location.state === null ? undefined : location.state
```
kjbekkelund: Was this because tests failed?

kjbekkelund:
Hi @justingreenberg, thanks for this! Only had a small question, otherwise I think this is good to go.

It would be great if you created a DevTools PR too!

src/index.js
justingreenberg: hi @kjbekkelund! yes, the tests were failing due to the way history assigns a `null` value to state, so this was the most lightweight way to keep state `undefined` internally

kjbekkelund: Ah, maybe we should change the tests instead? Nice to be consistent with `history`, plus the code is simpler.

jlongster:
:+1:

src/index.js
kjbekkelund: @justingreenberg If it isn't too much work changing the tests, maybe you could take a look when you've got the opportunity? No hurry — remember to relax and enjoy xmas :)

justingreenberg:
hey guys! @kjbekkelund ok so in :gift: 43be04e i removed the location.state expression and updated tests to expect state to be `null` after interacting with `history`... it's kinda awkward but it makes sense :)

let me know if you get around to reviewing, if not enjoy the holidays! :santa: :christmas_tree:

kjbekkelund:
@justingreenberg Nice, thanks! I'll try to get this tested and merged tomorrow. Great work :100:

Issue #140, opened 2015-12-27
Support ImmutableJS Map as the state


zavelevsky:
Although it makes sense to use ImmutableJS Map object as your state object - it currently can't be done because the reducer of redux-simple-router uses simple Object.assign.
Please consider supporting it by allowing to pass a custom reducer, or a reducer's internal state creator, such as (state, obj) => newState.

mxstbr:
:+1: I don't want to keep people from being able to use ImmutableJS if they want to with my boilerplate, would be nice if support for this added! (See https://github.com/mxstbr/react-boilerplate/issues/6)

neverfox:
+1 for a custom state access function. That seems to be the idiomatic way this is handled in other redux libraries that provide their own reducer.

jlongster:
We support a custom state accessor function in https://github.com/rackt/redux-simple-router#syncreduxandrouterhistory-store-selectrouterstate

neverfox:
Oh, ha. I just took the OP's issue at face value and didn't check that. I wonder what the problem is then. That should be all you need to support any immutable library.

jlongster:
I'm considering closing this unless the OP can describe further what's wrong. I know several people that use Immutable.js just fine with this.

zavelevsky:
Last I checked you return the result of Object.assign from the reducer.
This of course returns an object and not the original Map.

zavelevsky:
Original type, which is Map.

neverfox:
An original Map of what?

zavelevsky:
Sorry, I'm limited to answering by email at the moment.
As I recall, the lib's reducer returns the result of Object.assign,
therefore an object.
If my state is an ImmutableJS Map then after the reducer is called, what I
get is no longer an instance of a Map but a simple object.
Tjis might have changed since last I checked, but I cannot check at the
moment. Just wanted to clarify so the issue won't be closed prematurely.

neverfox:
So what you're asking for not just for the reducer to be able to get it's necessary sub-state out via a customer selector, but for the ability to tell the reducer how to write the new state back out, e.g. via `fromJS()`?

neverfox:
Could you not then just do something like:

```
syncReduxAndRouter(history, store, state => state.get('routing').toJS())

...

combindReducers({
  routing: (state, action) => routeReducer(state, action).fromJS()
})
```

zavelevsky:
Specifically, combineReducers doesn't support immutablejs for the same
[...]

Issue #122, opened 2015-12-16
@@router/INIT_PATH doesn't work with initalState?


theaqua:
Hi, I have this setup of my store:

```
const finalCreateStore = compose(
  autoRehydrate(),
  applyMiddleware(thunk, logger),
)(createStore);

const store = finalCreateStore(reducer, {
  routing: {
    path: `/contacts/30/about`,
  },
});
```

and then at `app.jsx` I use `syncReduxAndRouter(history, store)`.
When I start my app `redux-simple-router` ignores my `initialState` with `routing.path`. Why?

![localhost 3000 2015-12-16 04-03-43](https://cloud.githubusercontent.com/assets/775692/11828980/3632d884-a3aa-11e5-8b9e-7b16ececd8fa.png)

kjbekkelund:
Hi @theaqua, thanks for the input. I just don't think we ever thought about this use-case. I played around with this now, but got into a problem with our tests that makes this difficult right now.

So it's definitely possible to implement, but I think we need to solve the test issue before adding this (but we're very open to PRs, so if someone has an idea, please try it out and see if it works!)

@jlongster The problem as far as I can see is that `initialState` is "global state" across invocations of `syncReduxAndRouter`. My initial idea was to do something like this:

```javascript
if (!lastRoute) {
  // if getRouterState() is set, default to that
  // otherwise do what we do today
}
```

But that won't work in the tests as `getRouterState()` will only be "correct" on the first invocation of `syncReduxAndRouter` (so I got the feature working, but the tests failed, but they didn't fail "in the right way", as the test ran when I _only_ ran the failing test).

I think we need to consider moving everything into a function that you invoke (similar to `createHistory`), so we get an "instance" of redux-simple-router that you work with. I _think_ that might simplify some cases — but I really haven't gone through all the details, it just "feels" right. My next week will be busy, but I'll try to play around with ideas when I've got the time.

jlongster:
Yeah, it's not great that we change `initialState` from a single `syncReduxAndRouter`. Although I'd argue that you really can only sync once across your whole app. I'm open to other ways to set it up though.

This might be fixed by adding:

```js
  if(getRouterState().path) {
    lastRoute = getRouterState();
  }
```

Right when `syncReduxAndRouter` is called. But I'm not sure. I'm going through issues and trying to build up context for where we stand.

kjbekkelund:
Yep, no problem in apps, only a problem in the tests

timdorr:
[...]

Issue #138, opened 2015-12-25
Make redux drive history?


naw:
The idea behind `redux-simple-router` is to keep history in sync between redux and the router history.

In my opinion, it's preferable for redux to be the source of truth for the router, rather than history to be the source of truth that is shared by the router and redux.

My idea is this:

1. Let the browser history tell redux what to do (e.g. back, forward, initial page load, etc.)
2. Let links, buttons, etc. tell redux what to do (i.e. _not_ tell the router what to do).
3. Finally, let redux tell the router what to do.

I have a basic draft of this here:

https://github.com/naw/redux-simple-router/commit/5d6c893766ce2375f1d82c959fdfdc66a9f0ff2e

The basic implementation is:

1 Give react-router a memoryHistory object instead of the real browserHistory object.
2. The only thing that writes to memoryHistory is a redux store subscriber.
3. wrap the memoryHistory object we give to react-router so that we can intercept `push` calls so that they call a changeLocation action creator rather than writing to memoryHistory directly.
4. Separately, listen to a real browserHistory object, and call the changeLocation action creator from the history listener.

Also, I'm using a slightly different approach for preventing cycles, so I believe the `locationsAreEqual` complexity is unnecessary.

I've probably missed some edge cases (or perhaps I've missed something larger than that?), but I wanted to share the basic idea and get some feedback.

kjbekkelund:
Hi @naw, interesting approach. It's late, so I don't have to look much at this right now, so just a few questions that jumps to mind:

- Is `browserHistory.listen` guaranteed to be synchronously called?
- Have you ran it through all the tests? Lots of cases with regards to devtools etc
- Does this override `push` in `memoryHistory`? Seems "hacky" (though without looking much at this)

naw:
@kjbekkelund 

Thank you for your initial feedback.

The original commit I shared was a completely separate implementation providing a basic proof of concept by serving as a drop-in replacement within the example app.  It didn't actually rewrite the existing code, so none of the tests would apply.

Therefore, I have refactored my original implementation so that it is now an actual modification to the `redux-simple-router` code, keeping things as similar as possible (such as router names, action creator names, etc.) so that the tests can run with little modifications.

The new refactoring is here:

https://github.com/naw/redux-simple-router/commit/a47914fbb4b97a0cda1cebcb8df75b7246534e7a

Now to address your questions, and provide some additional clarification:

[...]

Issue #113, opened 2015-12-11
Redux DevTools issue (reapplying state)


fkrauthan:
It looks like there is still an issue if I go thru the list of state changes and disable a state everything is working fine (route is getting updated). But as soon as I enable that state again it looks like the browser url is not changing again. Let me know if you guys need more information.

fkrauthan:
It looks like it is not always happening. Only after a while when I play around with enabling and disabling different navigation states (e.g. navigating back and forth between two pages and then click on the DevTools to disable states and reenabling them.)

kjbekkelund:
Hm, interesting. I don't use the DevTools much, so I haven't seen this happen. Let me know if you find a reproducible way to make it fail. I'm probably going to focus on #108 and #109 first.

timdorr:
I think this is obsoleted by 2.0, but let me know if it's still happening and I can re-open.

PR #141, opened 2015-12-27
Use middleware to synchronize store to history

taion:
Per https://github.com/rackt/redux-simple-router/pull/129#issuecomment-167392040.

Fixes https://github.com/rackt/redux-simple-router/issues/108.

The idea is basically - keep unidirectional data flow for updates, most of the time. This works in most cases, but it breaks replay, since the actions in some sense aren't pure. That's fine, though - we can restrict the store observer to _only_ handle the replay case (and potentially not even use it otherwise).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,16 +78,21 @@ export function syncHistory(history) {
       const { location: initialLocation } = getRouterState()
 
       unsubscribeStore = store.subscribe(() => {
-        // If we're resetting to the beginning, use the saved values.
-        const storeLocation = getRouterState().location
-        const location = storeLocation || initialLocation
+        const { location } = getRouterState()
 
+        // If we're resetting to the beginning, use the saved initial value. We
+        // need to dispatch a new action at this point to populate the store
+        // appropriately.
+        if (!location) {
+          history.transitionTo(initialLocation)
+          return
+        }
+
+        // Otherwise, if we need to update the history location, do so without
+        // dispatching a new action, as we're just bringing history in sync
+        // with the store.
         if (location.key !== currentKey) {
-          // If there already is a location in the store, then don't dispatch
-          // an action from the history listener; otherwise do so in order to
-          // populate the store.
-          syncing = !!storeLocation
-
+          syncing = true
           history.transitionTo(location)
           syncing = false
         }
--- a/src/index.js
+++ b/src/index.js
@@ -1,107 +1,95 @@
 // Constants
 
-export const UPDATE_PATH = '@@router/UPDATE_PATH'
+export const TRANSITION = '@@router/TRANSITION'
+export const UPDATE_LOCATION = '@@router/UPDATE_LOCATION'
+
 const SELECT_STATE = state => state.routing
 
-export function pushPath(path, state, key) {
-  return {
-    type: UPDATE_PATH,
-    payload: { path, state, key, replace: false }
-  }
+function transition(method) {
+  return arg => ({
+    type: TRANSITION,
+    method, arg
+  })
 }
 
-export function replacePath(path, state, key) {
+export const push = transition('push')
+export const replace = transition('replace')
+
+// TODO: Add go, goBack, goForward.
+
+function updateLocation(location) {
   return {
-    type: UPDATE_PATH,
-    payload: { path, state, key, replace: true }
+    type: UPDATE_LOCATION,
+    location
   }
 }
 
 // Reducer
 
-let initialState = {
-  path: undefined,
-  state: undefined,
-  replace: false,
-  key: undefined
+const initialState = {
+  location: undefined
 }
 
-export function routeReducer(state=initialState, { type, payload }) {
-  if(type === UPDATE_PATH) {
-    return payload
+export function routeReducer(state = initialState, { type, location }) {
+  if (type !== UPDATE_LOCATION) {
+    return state
   }
 
-  return state
+  return { location }
 }
 
 // Syncing
-function createPath(location) {
-  const { pathname, search, hash } = location
-  let result = pathname
-  if (search)
-    result += search
-  if (hash)
-    result += hash
-  return result
-}
 
 export function syncHistory(history) {
   let unsubscribeHistory, currentKey, unsubscribeStore
-  let connected = false
+  let connected = false, syncing = false
 
   function middleware(store) {
     unsubscribeHistory = history.listen(location => {
-      const path = createPath(location)
-      const { state, key } = location
-      currentKey = key
+      currentKey = location.key
+      if (syncing) {
+        // Don't dispatch a new action if we're replaying location.
+        return
+      }
 
-      const method = location.action === 'REPLACE' ? replacePath : pushPath
-      store.dispatch(method(path, state, key))
+      store.dispatch(updateLocation(location))
     })
 
     connected = true
 
     return next => action => {
-      if (action.type !== UPDATE_PATH) {
+      if (action.type !== TRANSITION || !connected) {
         next(action)
         return
       }
 
-      const { payload } = action
-      if (payload.key || !connected) {
-        // Either this came from the history, or else we're not forwarding
-        // location actions to history.
-        next(action)
-        return
-      }
+      // FIXME: Is it correct to swallow the TRANSITION action here and replace
+      // it with UPDATE_LOCATION instead? We could also use the same type in
+      // both places instead and just set the location on the action.
 
-      const { replace, state, path } = payload
-      // FIXME: ???! `path` and `pathname` are _not_ synonymous.
-      const method = replace ? 'replaceState' : 'pushState'
-
-      history[method](state, path)
+      const { method, arg } = action
+      history[method](arg)
     }
   }
 
   middleware.syncHistoryToStore =
     (store, selectRouterState = SELECT_STATE) => {
       const getRouterState = () => selectRouterState(store.getState())
-      const {
-        key: initialKey, state: initialState, path: initialPath
-      } = getRouterState()
+      const { location: initialLocation } = getRouterState()
 
       unsubscribeStore = store.subscribe(() => {
-        let { key, state, path } = getRouterState()
-
         // If we're resetting to the beginning, use the saved values.
-        if (key === undefined) {
-          key = initialKey
-          state = initialState
-          path = initialPath
-        }
+        const storeLocation = getRouterState().location
+        const location = storeLocation || initialLocation
 
-        if (key !== currentKey) {
-          history.pushState(state, path)
+        if (location.key !== currentKey) {
+          // If there already is a location in the store, then don't dispatch
+          // an action from the history listener; otherwise do so in order to
+          // populate the store.
+          syncing = !!storeLocation
+
+          history.transitionTo(location)
+          syncing = false
         }
       })
     }
--- a/src/index.js
+++ b/src/index.js
@@ -1,61 +1,40 @@
-import deepEqual from 'deep-equal'
-
 // Constants
 
 export const UPDATE_PATH = '@@router/UPDATE_PATH'
 const SELECT_STATE = state => state.routing
 
-export function pushPath(path, state, { avoidRouterUpdate = false } = {}) {
+export function pushPath(path, state, key) {
   return {
     type: UPDATE_PATH,
-    payload: {
-      path: path,
-      state: state,
-      replace: false,
-      avoidRouterUpdate: !!avoidRouterUpdate
-    }
+    payload: { path, state, key, replace: false }
   }
 }
 
-export function replacePath(path, state, { avoidRouterUpdate = false } = {}) {
+export function replacePath(path, state, key) {
   return {
     type: UPDATE_PATH,
-    payload: {
-      path: path,
-      state: state,
-      replace: true,
-      avoidRouterUpdate: !!avoidRouterUpdate
-    }
+    payload: { path, state, key, replace: true }
   }
 }
 
 // Reducer
 
 let initialState = {
-  changeId: 1,
   path: undefined,
   state: undefined,
-  replace: false
+  replace: false,
+  key: undefined
 }
 
-function update(state=initialState, { type, payload }) {
+export function routeReducer(state=initialState, { type, payload }) {
   if(type === UPDATE_PATH) {
-    return Object.assign({}, state, {
-      path: payload.path,
-      changeId: state.changeId + (payload.avoidRouterUpdate ? 0 : 1),
-      state: payload.state,
-      replace: payload.replace
-    })
+    return payload
   }
+
   return state
 }
 
 // Syncing
-
-function locationsAreEqual(a, b) {
-  return a != null && b != null && a.path === b.path && deepEqual(a.state, b.state)
-}
-
 function createPath(location) {
   const { pathname, search, hash } = location
   let result = pathname
@@ -66,84 +45,75 @@ function createPath(location) {
   return result
 }
 
-export function syncReduxAndRouter(history, store, selectRouterState = SELECT_STATE) {
-  const getRouterState = () => selectRouterState(store.getState())
+export function syncHistory(history) {
+  let unsubscribeHistory, currentKey, unsubscribeStore
+  let connected = false
 
-  // To properly handle store updates we need to track the last route.
-  // This route contains a `changeId` which is updated on every
-  // `pushPath` and `replacePath`. If this id changes we always
-  // trigger a history update. However, if the id does not change, we
-  // check if the location has changed, and if it is we trigger a
-  // history update. It's possible for this to happen when something
-  // reloads the entire app state such as redux devtools.
-  let lastRoute = undefined
+  function middleware(store) {
+    unsubscribeHistory = history.listen(location => {
+      const path = createPath(location)
+      const { state, key } = location
+      currentKey = key
 
-  if(!getRouterState()) {
-    throw new Error(
-      'Cannot sync router: route state does not exist (`state.routing` by default). ' +
-      'Did you install the routing reducer?'
-    )
-  }
+      const method = location.action === 'REPLACE' ? replacePath : pushPath
+      store.dispatch(method(path, state, key))
+    })
 
-  const unsubscribeHistory = history.listen(location => {
-    const route = {
-      path: createPath(location),
-      state: location.state
-    }
+    connected = true
 
-    if (!lastRoute) {
-      // `initialState` *should* represent the current location when
-      // the app loads, but we cannot get the current location when it
-      // is defined. What happens is `history.listen` is called
-      // immediately when it is registered, and it updates the app
-      // state with an UPDATE_PATH action. This causes problem when
-      // users are listening to UPDATE_PATH actions just for
-      // *changes*, and with redux devtools because "revert" will use
-      // `initialState` and it won't revert to the original URL.
-      // Instead, we specialize the first route notification and do
-      // different things based on it.
-      initialState = {
-        changeId: 1,
-        path: route.path,
-        state: route.state,
-        replace: false
+    return next => action => {
+      if (action.type !== UPDATE_PATH) {
+        next(action)
+        return
       }
 
-      // Also set `lastRoute` so that the store subscriber doesn't
-      // trigger an unnecessary `pushState` on load
-      lastRoute = initialState
+      const { payload } = action
+      if (payload.key || !connected) {
+        // Either this came from the history, or else we're not forwarding
+        // location actions to history.
+        next(action)
+        return
+      }
 
-      store.dispatch(pushPath(route.path, route.state, { avoidRouterUpdate: true }));
-    } else if(!locationsAreEqual(getRouterState(), route)) {
-      // The above check avoids dispatching an action if the store is
-      // already up-to-date
-      const method = location.action === 'REPLACE' ? replacePath : pushPath
-      store.dispatch(method(route.path, route.state, { avoidRouterUpdate: true }))
+      const { replace, state, path } = payload
+      // FIXME: ???! `path` and `pathname` are _not_ synonymous.
+      const method = replace ? 'replaceState' : 'pushState'
+
+      history[method](state, path)
     }
-  })
+  }
 
-  const unsubscribeStore = store.subscribe(() => {
-    let routing = getRouterState()
+  middleware.syncHistoryToStore =
+    (store, selectRouterState = SELECT_STATE) => {
+      const getRouterState = () => selectRouterState(store.getState())
+      const {
+        key: initialKey, state: initialState, path: initialPath
+      } = getRouterState()
 
-    // Only trigger history update if this is a new change or the
-    // location has changed.
-    if(lastRoute.changeId !== routing.changeId ||
-       !locationsAreEqual(lastRoute, routing)) {
+      unsubscribeStore = store.subscribe(() => {
+        let { key, state, path } = getRouterState()
 
-      lastRoute = routing
-      const method = routing.replace ? 'replace' : 'push'
-      history[method]({ 
-        pathname: routing.path, 
-        state: routing.state 
+        // If we're resetting to the beginning, use the saved values.
+        if (key === undefined) {
+          key = initialKey
+          state = initialState
+          path = initialPath
+        }
+
+        if (key !== currentKey) {
+          history.pushState(state, path)
+        }
       })
     }
 
-  })
-
-  return function unsubscribe() {
+  middleware.unsubscribe = () => {
     unsubscribeHistory()
-    unsubscribeStore()
-  }
-}
+    if (unsubscribeStore) {
+      unsubscribeStore()
+    }
 
-export { update as routeReducer }
+    connected = false
+  }
+
+  return middleware
+}
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,3 @@
 {
-    "presets": ["es2015"],
-    "plugins": ["transform-object-assign"]
+    "presets": ["es2015", "stage-2"]
 }
--- a/package.json
+++ b/package.json
@@ -42,8 +42,8 @@
     "babel-core": "^6.2.1",
     "babel-eslint": "^4.1.6",
     "babel-loader": "^6.2.0",
-    "babel-plugin-transform-object-assign": "^6.0.14",
     "babel-preset-es2015": "^6.1.2",
+    "babel-preset-stage-2": "^6.3.13",
     "eslint": "^1.10.3",
     "eslint-config-rackt": "^1.1.1",
     "expect": "^1.13.0",
@@ -64,8 +64,5 @@
     "redux": "^3.0.4",
     "redux-devtools": "^2.1.5",
     "webpack": "^1.12.9"
-  },
-  "dependencies": {
-    "deep-equal": "^1.0.1"
   }
 }
--- a/test/createTests.js
+++ b/test/createTests.js
@@ -1,8 +1,8 @@
 /*eslint-env mocha */
 
 import expect from 'expect'
-import { pushPath, replacePath, UPDATE_PATH, routeReducer, syncReduxAndRouter } from '../src/index'
-import { createStore, combineReducers, compose } from 'redux'
+import { pushPath, replacePath, UPDATE_PATH, routeReducer, syncHistory } from '../src/index'
+import { applyMiddleware, createStore, combineReducers, compose } from 'redux'
 import { devTools } from 'redux-devtools'
 import { ActionCreators } from 'redux-devtools/lib/devTools'
 import { useBasename } from 'history'
@@ -11,27 +11,26 @@ expect.extend({
   toContainRoute({
     path,
     state = undefined,
-    replace = false,
-    changeId = undefined
+    replace = false
   }) {
     const routing = this.actual.getState().routing
 
     expect(routing.path).toEqual(path)
     expect(routing.state).toEqual(state)
     expect(routing.replace).toEqual(replace)
-
-    if (changeId !== undefined) {
-      expect(routing.changeId).toEqual(changeId)
-    }
   }
 })
 
 function createSyncedHistoryAndStore(createHistory) {
-  const store = createStore(combineReducers({
+  const history = createHistory()
+  const middleware = syncHistory(history)
+  const { unsubscribe } = middleware
+
+  const createStoreWithMiddleware = applyMiddleware(middleware)(createStore)
+  const store = createStoreWithMiddleware(combineReducers({
     routing: routeReducer
   }))
-  const history = createHistory()
-  const unsubscribe = syncReduxAndRouter(history, store)
+
   return { history, store, unsubscribe }
 }
 
@@ -50,17 +49,17 @@ module.exports = function createTests(createHistory, name, reset = defaultReset)
             path: '/foo',
             replace: false,
             state: { bar: 'baz' },
-            avoidRouterUpdate: false
+            key: undefined
           }
         })
 
-        expect(pushPath('/foo', undefined, { avoidRouterUpdate: true })).toEqual({
+        expect(pushPath('/foo', undefined, 'foo')).toEqual({
           type: UPDATE_PATH,
           payload: {
             path: '/foo',
             state: undefined,
             replace: false,
-            avoidRouterUpdate: true
+            key: 'foo'
           }
         })
       })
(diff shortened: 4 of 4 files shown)


taion:
This actually fixes https://github.com/rackt/redux-simple-router/issues/140 as well.

taion:
I believe this fixes https://github.com/rackt/redux-simple-router/issues/122, too.

taion:
Probably also https://github.com/rackt/redux-simple-router/issues/113, since I think using `location.key` avoids the edge cases you get from an incrementing `changeId`.

taion:
Note that you cannot in general directly update the store based on just the action - it's possible for e.g. a `listenBefore` hook to block the transition, and that would then take the store out-of-sync with the browser.

tomatau:
Just tried it out and it does fix https://github.com/rackt/redux-simple-router/issues/108 with a nice implementation. Feels more suitable to redux as a simple middleware.

I think it would be worth updating usage and example in this PR for clarity.

ezekielchentnik:
@taion I believe location.key is not unique, it's random - different every time .push/replace are called, regardless if same path, state.  not sure if that matters...

taion:
I know exactly what `location.key` is, thanks.

taion:
@tomatau I see you beat me to it. The one wart is that you need `middleware.syncHistoryToStore` per https://github.com/rackt/redux-simple-router/pull/141/files#diff-413f2628e9c93c46cec6db202ddb8ea4R159 if you want the Redux DevTools to work (but in normal operation that bit is a no-op).

Wanted to get some feedback from @jlongster and @kjbekkelund as to whether this approach seems workable. I don't use Redux at all myself, but it seems like preserving unidirectional data flow via a transducer just gets rid of all the cycle management headaches (though it requires that extra line of boilerplate for DevTool support).

This is halfway a PoC just to illustrate what's involved though. Ideally, for a breaking change, it'd be best to make the user-dispatched actions take location descriptors (and correspondingly probably just have a different `type`), while having the reducer operate on actual locations from `history`.

kjbekkelund:
@taion Very interesting! I'm quite busy until early January, but hopefully @jlongster has got some time to review and test. Otherwise I'll test as soon as I've got some spare hours :)

tomatau:
> The one wart is that you need middleware.syncHistoryToStore per https://github.com/rackt/redux-simple-router/pull/141/files#diff-413f2628e9c93c46cec6db202ddb8ea4R159 if you want the Redux DevTools to work (but in normal operation that bit is a no-op).

Testing this out seems buggy as it's dispatching new updatePath actions when rolling back history... which is very bad and counter-intuitive for debugging. Not adding this line seems like a nicer experience, to just go without reversing through the path history from DevTools rollbacks.

I appreciate that this is mainly a PoC, but I'm very keen to see one of these new implementations get merged and fix up https://github.com/rackt/redux-simple-router/issues/108 ASAP :) So put the word out for any work to get movement and I'm happy to contribute.

taion:
It's tricky. Suppose you want to reverse the path history.

This necessarily has to have a side effect. I think the right way to model this on the `history` side is with a `push`, but that will create a new `location` with a new `location.key`.

If I don't dispatch the extra "update path" action, then the `key` for the `location` would be out of sync between `history` and `store`.

It's possible to directly use `history.transitionTo` and keep the same `key`, but that means we'd want to preserve the action involved, and it'd seem odd to me to issue a `REPLACE` or a `POP` to the `history` rather than a `PUSH` here.

taion:
Well, maybe using `history.transitionTo` would be all right.

jlongster:
I've been resisting to forcing a middleware on the user, as I think a middleware should only be for converting one action into multiple actions, but I haven't had time to look closely at this. I will do so this week though.

Also, long-term you and other people who work on react-router will probably be involved in maintaining this so you have a greater say in how this evolves than other PRs in my opinion.

taion:
I think the problem with round-tripping this through the store is that it breaks unidirectional data flow. If you do this before it hits the store in the first place, and only use the store for restoring location on replay, you completely exclude issues like #108; otherwise it just feels like a bunch of work to avoid cascading updates, which Flux-like architectures should inherently exclude.

jlongster:
Yeah, there are benefits to it, but I'd say that #129 is still not that much work and it avoids other problems that this may introduce (there are so many little edge cases with the devtools), without the need for the user to install a middleware. I'm open to it! Just need to look at it more.

taion:
Let me flesh this out with the changes we mentioned earlier.

taion:
Updated the PR. This now closes https://github.com/rackt/redux-simple-router/issues/95 as well. It's a more significant API change, though.

taion:
@tomatau 

This should fix your https://github.com/rackt/redux-simple-router/pull/141#issuecomment-167618037, except when completely clearing all the actions.

The idea is that this adds the same asymmetry between providing a location descriptor to the action creators, while populating the store with fully fleshed-out location objects.

jlongster:
@taion I've always wanted to avoid forcing the user to install a middleware. However, I don't really care that much. Really, the two core ideas that need to exist for me to be happy are: only dumb JS objects stored in the app state, and the user uses react-router's API directly in routing components (and wherever else they. As long as that's the case, I don't really care how it works.

You have a much better understanding of how more complex stuff like async routing works also.

If we pursue this path, are you going to continually be involved here? I foresee myself being less involved over time, as I was really hoping to kick off this new direction (and I'm happy it succeeded). I'd love it if you and other react-router folks eventually took over maintenance, but I don't know how burdened you all are with that already.

src/index.js
```diff
@@ -1,149 +1,107 @@
...
+  })
 }
 
-export function replacePath(path, state, { avoidRouterUpdate = false } = {}) {
+export const push = transition('push')
+export const replace = transition('replace')
+
+// TODO: Add go, goBack, goForward.
+
+function updateLocation(location) {
```
jlongster: I'm assuming this `location` object is now the fully parsed location? Can we make it so you can also pass just a string, and if you do that we just convert it into `{ pathname: location }`?

jlongster: I know that makes the API slightly asymmetrical (the state would be different than what you passed in), but now that we can always store the same shape of object in the state I think it's ok.

taion: It's already supported. See e.g. https://github.com/rackt/redux-simple-router/pull/141/files#diff-413f2628e9c93c46cec6db202ddb8ea4R338.

taion: Oops - misread. I'm actually not exporting `updateLocation` at all. The `transition` actions do take strings, though.

src/index.js
```diff
@@ -1,149 +1,107 @@
...
-  }
+function transition(method) {
+  return arg => ({
+    type: TRANSITION,
+    method, arg
+  })
 }
 
-export function replacePath(path, state, { avoidRouterUpdate = false } = {}) {
+export const push = transition('push')
```
jlongster: `push` and `replace` are *very* vague names when seen outside of react-router. Would you be ok keeping the Path suffix? I think it generally makes it clearer.

src/index.js
jlongster: oh, great, looks like that's a natively-supported thing in react-router.

src/index.js
taion: I agree that `push` and `replace` are awful names. Maybe `pushLocation` and `replaceLocation`? I thought `pushPath` was a bit inaccurate, since this would let you push `query`, &c. as well. I also didn't want to figure out how to then name `go`, `goBack`, `goForward` – would we suffix those with `location` as well?

src/index.js
taion: Yes – from the `TRANSITION` actions. This is just for receiving the update from `history` for the reducer.

taion:
This doesn't touch on async routing at all – and it can't, without a much larger re-write. The routing data is only exposed to the children of `<Router>` right now. By contrast, as you've discussed elsewhere, the library as-is only integrates into `history`. It's difficult; I can see some ways to make it work, but since I don't actually use Redux, I'd have no clue whether the API would be satisfactory to users or not.

As for my time involvement, I don't know – I don't use Redux, so this is just my vague impressionistic idea of how a Redux integration might look from what I've seen on paper.

Maybe to break down the pieces here a little bit, though, there are a few things here worth highlighting:

- I feel like the primary purpose of an integration like this is to sync history and store in normal use. Intercepting everything in a middleware gives you that, because the data flow is just either action -> history -> store, or history -> store; you don't have to deal with both an action -> store -> history path and a separate history -> store path.
  - Correspondingly, the dev tools integration to _cause_ side effects on modifying action history is secondary; if the user doesn't want this, he or she doesn't have to use `middleware.syncHistoryToStore` at all, so it's the same amount of API work – just a matter of installing the middleware instead of syncing the `history` to the `store`
- With the new history APIs, the natural point of integration is having the user specify a location descriptor, but putting location objects in the store. Among other things, this is how all the `history` APIs will work, and it can allow e.g. supporting named routes in a very generic way (`history.push({ name: 'foo', params })`) and have things _just work_ for a user of this library. This means, though, that the "raw" user actions probably shouldn't hit the reducer without going through `history` first, since you want locations at the reducer, not location descriptors.
- Using `changeId` seems quite fragile – you're probably better off using `location.key` instead, and using `history.transitionTo` to restore the key after a dev tools action.

taion:
To elaborate, with location descriptors, you really want to separate the location descriptor the user provides from the location that goes into the store. That means the current approach of feeding the user-supplied action directly into the reducer is unlikely to work well.

You could e.g. have a "last location descriptor action" reducer that a store subscriber listens to, then uses to update `history`, and you'd still have the same architectural design where all updates to the actual reducer for the locations comes from the `history` (necessarily so), but I just think a middleware is cleaner (really, more unidirectional) here.

jlongster:
Yeah, I think you've convinced me. The big win is that what is stored in the state is a consistent location object coming from history, no matter what the user called the action creator with. This also opens up the path to supporting routing params if we want to do that.

Note that #129 does not use `changeId`, it's a different implementation and you should look at that for the "latest" ideas.

taion:
Just updated the PR again to fix an edge case.

What do you think of the middleware swallowing the `TRANSITION` action? I set it up that way to avoid having to change the `'only triggers store once when updating path via store'` test, but maybe that's not so correct - the default reducer already ignores the `TRANSITION` action anyway.

jlongster:
What happens in that test differently now? Why is the store updated more than once?

Reading your comment: "We could also use the same type in both places" How could we use the same action type, wouldn't that be an infinite loop because it keeps triggering the middleware?

taion:
1. If I don't swallow the `TRANSITION` action, I see the store subscriber firing on both the `TRANSITION` action (for a no-op non-change) and the `UPDATE_LOCATION` action (that does cause the store to update)
2. I probably should have added a test case - just tracing through the logic on paper, if you reset all the actions while on the initial location, I think my implementation per https://github.com/rackt/redux-simple-router/commit/243235eb7c1c9c5c79ab22e2e1eddc88123357d2 would have left the store empty, since we would have had `location.key === currentKey`. So in the special case where the store is cleared, we need to make `history` cause an `UPDATE_LOCATION` action to be dispatched regardless.
3. This is probably really messy, but you could do something along the lines of e.g. https://github.com/rackt/redux-simple-router/commit/6fc6ab7a5fd29cf8cb7ab0f39a528a10cdc30d5f#diff-1fdf421c05c1140f6d71444ea2b27638R71, where actions that don't have e.g. a `location` key just get swallowed and sent to `history` instead (but ones with a `location` key get passed through). This is probably as horrible/hacky as it sounds, right?

jlongster:
> If I don't swallow the TRANSITION action, I see the store subscriber firing on both the TRANSITION action (for a no-op non-change) and the UPDATE_LOCATION action (that does cause the store to update)

Ah, right.

> I probably should have added a test case - just tracing through the logic on paper, if you reset all the actions while on the initial location, I think my implementation per 243235e would have left the store empty, since we would have had location.key === currentKey. So in the special case where the store is cleared, we need to make history cause an UPDATE_LOCATION action to be dispatched regardless.

Yeah, clearing all actions is a weird edge case to support from the redux devtools. Mainly because the initial routing state is actually wrong and it requires one action during init to be in sync with the current location.

It sounds like you've been able to figure it out though.

> This is probably really messy, but you could do something along the lines of e.g. 6fc6ab7#diff-1fdf421c05c1140f6d71444ea2b27638R71, where actions that don't have e.g. a location key just get swallowed and sent to history instead (but ones with a location key get passed through). This is probably as horrible/hacky as it sounds, right?

Yeah, I think I like having two action types better. It's clearer and easier to debug.

Redux devtools won't see any of the `TRANSITION` actions, but only `UPDATE_LOCATION`, right?

src/index.js
jlongster: Hm that's a good point. `goPath` doesn't really sound great. We could just rename `push` and `replace` and leave the other be, but if you think that's confusing we can just stick with these simple names and users can rename them when importing.

jlongster:
I'll actually use this tomorrow (or, worst case later this week), but if it seems to work well I'm tempted to just merge it. We need to settle on something, and this sounds like the most promising solution so far.

src/index.js
taion: I don't know the naming conventions for Redux action creators at all. Is there an option to namescope these under e.g. `HistoryActionCreators` or something, like it's done with the devtools action creators?

taion:
Right now the `TRANSITION` actions get swallowed. It won't mess with anything on the history side (or w/devtools) to let them through, but I don't know if that would lead to extra re-rendering or something, even if the store isn't actually updated at all by the reducers.

jlongster:
Right, but it's a middleware that swallows it. If any middleware is run before it, it will see the the action. It's common in redux to specify the order you must list middleware, so I just want to make sure we say something about it if we need to. Like "make sure you put this before devtoolsMiddleware()"

taion:
I thought the dev tools (1) weren't a middleware and (2) always go at the end anyway? But yeah, this should definitely go before the dev tools.

jlongster:
Ah, it's not a middleware, it's a "store enhancer" like `applyMiddleware`. So we don't need to explain anything in the docs, it's already common practice that middleware gets executed first.

jlongster:
I like it. Let's merge it and just see what else we need to fix up for a new version. (new docs obviously)

timdorr:
If you need any help with this, I actually *do* use Redux and Router regularly. Though I haven't use a binding library (yet), I can look at anything from the perspective of The Redux Way™ if you need.

taion:
I think the biggest open issue is action creator naming, per https://github.com/rackt/redux-simple-router/pull/141/files#r48519831 and follow-ups.

timdorr:
An entirely unofficial, but popular, format for actions that many libraries use is a Flux Standard Action: https://github.com/acdlite/flux-standard-action

You might try putting information on the meta field. That is where I'm inclined to investigate. 

But let me check this out and integrate it into my code base to really get a feel for it. Is this pushed up to NPM?

taion:
I don't mean the format (though we can namespace everything back under `payload` again if we want). I actually mean what to call the action creator functions. `push` and `replace` are too generic by a lot.

src/index.js
timdorr: This could work:
```js
import { routeActions } from 'redux-simple-router'
routeActions.push('/foo')
```
It would also make using `bindActionCreators` easier because they're all under the same object namespace.

timdorr: Case in point:
```js
@connect(
  someSelector,
  dispatch => ({ router: bindActionCreators({ routeActions }, dispatch) })
)
class FooBar extends Component {
  handleClick = () => {
    this.props.router.push("/foo")
  }
}
```
I like that a lot.

taion: PR it then :p

timdorr: Done! #149

neverfox:
Great work on all of this! Sounds like my `goBack` button will now work without having to be pressed twice. ;)

What's the currently release plan for all of the new goodies on master?

gaearon:
Can somebody help me understand why `syncHistoryToStore` is necessary? I haven't seen methods on middleware before and it looks like boilerplate. Ideally passing the middleware should be enough for it to run. What am I missing?

taion:
@gaearon It's for the dev tools. It's to enable the feature where, if the users manipulates the sequence of history-related actions, we sync the current history location to what's in the store.

gaearon:
Yeah, but why is it separate? Is it incurring a production cost? Should it explicitly mention devtools? Right now the API makes it seem like it's necessary to call.

taion:
@gaearon 

What do you mean "why is it separate"? My impression is that I can't do this within just the middleware itself, since the middleware runs before the dev tools.

Is there a more appropriate idiom here that would both let me transform the action before it hits the dev tools, and also subscribe to changes in the store?

One benefit, though, to setting this up this way, is that only the store subscriber cares about where the routing state ends up in the store - the middleware itself only works on the level of the actions.

gaearon:
>What do you mean "why is it separate"? My impression is that I can't do this within just the middleware itself, since the middleware runs before the dev tools.

Not necessarily:

```
compose(
  applyMiddleware(a, b),
  DevTools.instrument(),
  applyMiddleware(c, d)
)
```

But unfortunately I don't understand the problem well enough to answer your question. If you have time, I'd appreciate a detailed writeup of what happens with and without that call, using some specific example (user went there, user pressed that, etc).

taion:
I'd like to understand what the correct patterns here are. Right now, the code works like this:

1. User dispatches a `TRANSITION` action
2. The middleware listens to this `TRANSITION` action – it then calls e.g. `history.push` based on the contents of the `TRANSITION` action
3. `history` updates with the new location and calls its subscribers
  - **NOTE:** This is also the entry point for transitions triggered by e.g. `history.push` or a React Router `<Link>` directly, i.e. the ones that don't go through Redux actions
4. The `history` subscriber installed by the middleware dispatches an `UPDATE_LOCATION` action with the new location
5. The middleware in (2) doesn't do anything to this action
6. The dev tools observe this action and record it (as a staged action or something?)
7. Schematically, the relevant reducer handles this action to make the location available to store observers

Now the question is what happens when the developer cancels this action in the dev tools.

My understanding is that in most cases, cancelling an action should have no further side effects – it's sufficient for the dev tools to just update the store.

However, in some cases, when working with `history`, it might be desirable to actually also cause a side effect of transitioning to the previous location when cancelling `UPDATE_LOCATION` actions, e.g. from (4) above.

The way that `syncHistoryToStore` effects this (and only this) is to install a subscriber on the store. When it sees the store update with a location that wasn't the one most recently published by the `history` as in (3) above, it will call into the history to sync the location in history with the one in the store.

This might just be my lack of understanding, but it seemed like subscribing to the store for updates caused by the user e.g. cancelling actions in the dev tools was the best way to handle this, because at a glance it looked like the dev tools was publishing its calculated store state, rather than replaying some set of actions.

gaearon:
If you want to subscribe to the store (and it seems that you do), the “Redux way” of doing that would be to turn this middleware into a store enhancer (sounds like Redux Router :-).

taion:
Makes sense. The other two thoughts were:

1. This thing is a weird special-case side effect, because I only care about the terminal state the store ends up in; I don't want to e.g. replay all the `UPDATE_LOCATION` actions on the `history`
2. Maybe you don't always want to sync history to the store – the middleware works only on the level of the actions, whereas subscribing to the store seems to require me to make assumptions as to how the reducer is installed, which seems a bit more involved

gaearon:
Can there be two different middlewares for that?

```js
applyMiddleware(reduxSimpleRouter),
DevTools.instrument(),
applyMiddleware(syncHistoryWithDevTools)
```

?

taion:
Actually, looking at this a bit more, do you think it might be better to make this a dev tools monitor instead? `<SyncHistoryMonitor>` or something.

gaearon:
FWIW I'd release a version that doesn't sync the URL back first. I'm not convinced it's that useful to perform side effect in this case. Then we can see if there's enough demand for it and consider different options.

taion:
I'm totally on board with that. @tomatau mentioned it earlier in https://github.com/rackt/redux-simple-router/pull/141#issuecomment-167618037 as well, though in reference to an earlier, buggier implementation. It makes things a lot messier, and I have no idea what the workflow benefit actually is.

It's worth noting that this is exactly what you get right now if you just don't call `middleware.syncHistoryToStore`, though, and intentionally so, since this did seem very secondary to me.

I just didn't feel inclined to delete the whole "dev tools compatibility" section of the test suite.

gaearon:
I think "compatibility" means "keeps working predictably regardless of what's in URL bar". As long as it doesn't crash with DevTools open and lets you go back in time, with the app updating visually to render previous routes, I think it's compatible. 

Now if the app doesn't update to render the previous routes, that would be a problem. I'm not sure what is considered the source of truth for the location React Router sees. Can we make the app update visually when you toggle history actions (and screw URL bar)?

taion:
Not easily – you'd need something like https://github.com/rackt/redux-simple-router/issues/138, where Redux actually controls the `history` that the router uses as its source of truth for the current location, or at least a more restricted version that injects rewinded location states into the router.

I think that takes us back to redux-router-like levels of complexity.

But I think the complexity is just about equivalent with this implementation – you just call into `history.transitionTo` again, and you end up where you want to be with respect to the router.

Ultimately the issue here is that, since Redux isn't driving the router state, we need to do _something_ impure after dev tools actions to sync the router state back up to what's in the store. I think something like the current approach with an explicit "sync" makes the non-dev-tools code path a lot simpler and more predictable – we just need to be a bit careful with edge case handling on syncing the history after manipulating action history.

Whether that should be a store enhancer or whatever – that's just shuffling around a few lines of code; I think the basic logic right now is essentially correct.

ezekielchentnik:
@gaearon I'm curious as to when it's appropriate to create a store enhancer. I understand the differences between middleware and store enhancers, but is there some guidance as to when it is a good or bad idea to create a store enhancer vs handling functionality externally?

tomatau:
@taion & @gaearon 

Working with `syncHistoryToStore` is still not intuitive:
 * when rolling back, the app state rolls as expected but the path doesn't change (for me)
 * if actions are tied to routes, new actions can be dispatched when working with the dev tools - this doesn't align with the idea of "replaying state".

So, I still feel that `syncHistoryToStore` should be removed from `redux-simple-router`.  Both the ideas of making a new "SimpleRouterHistoryMonitor" or another middleware for this purpose seem sensible.

@taion - the internal complexity of redux-router was nowhere near as big a turn off as the complexity of initial setup for consumers. Redux-simple-router will never reach the same level of API complexity, so I think we shouldn't be putting too much weight on internal complexity when deciding an implementation.

Surely, the aim should be for the most well integrated solution, and so far, from the different approaches I've tested out, having redux drive the history updates using a history wrapper was the best for that... On the other hand, this middleware approach feels the most "reduxy" and least imposing. So for me, both of these approaches hold similar weight.

I'd quite like to see the "redux drive history" approach to be explored further, as it seems it was shot down prematurely due to worries of internal complexity, where really, it was doing quite good job of keeping state and history in sync. History methods just worked. DevTools just worked.

If now is the time to decide how this library will be implemented, it would be worth giving all the solutions a good kicking around and then we can all have more confidence that the best solution was chosen.

jlongster:
@tomatau If rolling back actions does not change the URL, please give detailed steps to reproduce. When I last tried this, it worked for me. And the whole point of `transitionTo` inside the store subscriber was to change the URL without dispatching actions.

I think we definitely have room to play around with things, I'm more open to things like middleware/store enhancers now and we aren't near the complexity that redux router was. It had some good ideas though. I do think we should avoid wrapping history, etc. We can make everything work beautifully without too much work, but getting devtools to work is the most complicated, and from what I've heard redux-router didn't work great with it as well.

> FWIW I'd release a version that doesn't sync the URL back first. I'm not convinced it's that useful to perform side effect in this case. Then we can see if there's enough demand for it and consider different options.

If by syncing URL you mean to rerender the route when actions are rolled back, that's an often-requested feature and many of the bugs that come in are from people using devtools and expected it to "just work".

mjrussell:
@tomatau @jlongster the path not changing is probably due to #164, I was playing with it a lot after spending some time upgrading/slimming down redux-router (following React-Router's v2) and running across similar history replace issues and was curious how redux-simple-router was implementing them.

It only happens when hitting the POP's (in this case resetting an action that directly follows a POP) because it cannot correctly replay the POP event when it is adding to the store.

mjrussell:
I tried writing a breaking test for it, but trigger a POP event is actually pretty hard unless you have full control over the browser and the tests are currently designed to run without knowledge of the browser (which works great in every other case)

taion:
@mjrussell 

Just use `createMemoryHistory` and call `goBack` or whatever. That said, I just don't know how to deal with this right now.

The problem with the "Redux drives history" approach is that, right now, `history` doesn't even drive history – it's just a proxy for the browser history API.

I don't think in a production use case, you want to deal with trying to completely hijack the browser history API. That API isn't designed for this, and it doesn't really play nicely with this sort of control.

I'm fine with complexity - we actually have a ton of complexity in the RR and history RCs for handling backward compat. What I'm not okay with is flakiness, and I think hijacking the browser history APIs is going to lead to unacceptable flakiness for production, user-facing pages.

That said, "Redux drives history" might be a decent approach for playing nicely with dev tools...

tomatau:
> @tomatau If rolling back actions does not change the URL, please give detailed steps to reproduce. 

@jlongster I have a repo branch that you can see the behaviour in action. [here](https://github.com/tomatau/breko-hub/tree/rSrTest)

This is the syncHistoryToStore call on the client https://github.com/tomatau/breko-hub/blob/rSrTest/src/app/entry.js#L23

And here is where the middleware is added on the client  https://github.com/tomatau/breko-hub/blob/rSrTest/src/app/state/middleware.js#L17

It will sporadically update the path when routes are updated through the DevTools, but 9/10 times, the path doesn't update. Also, when the path does update via the DevTools, it's coming from a new action, which it shouldn't be.

This is on `"redux-simple-router": "^2.0.1"`

taion:
Can you step through ~https://github.com/rackt/redux-simple-router/blob/master/src/index.js#L84-L102 to see what's happening? Unless you're completely wiping history (in which case we _need_ to push a new action to get the store into the correct state), you should hit the forward branch there, and as long as `syncing` is set, the history shouldn't update the store.

jlongster:
Do you all mind looking through the docs here: https://github.com/rackt/redux-simple-router/blob/master/README.md particularly the API section? Is everything accurate?

I will push out 2.0.0 today or tomorrow.

taion:
- The middleware has an `unsubscribe` method that unsubscribes everything
- The reducer is only necessary if you're using `syncHistoryToStore` (or if you want to, you know, actually observe the location); the middleware itself will work without it otherwise
- `goForward` and `goBack` have no args, while `go` just takes a `n` (signed int)

timdorr:
>  -The middleware has an unsubscribe method that unsubscribes everything

Fixed!

> - The reducer is only necessary if you're using `syncHistoryToStore` (or if you want to, you know, actually observe the location); the middleware itself will work without it otherwise

Correct, but I wouldn't want to lead beginners down the path of thinking it's optional and then having them be confused when `state.routing` is empty.

> - `goForward` and `goBack` have no args, while `go` just takes a `n` (signed int)

Fixed!

billyjanitsch:
Sorry that I'm commenting after v2 has shipped, but for what it's worth, I think this new API has bad implications for separation of concerns, and is a lot more complicated to setup.

The example given in the [tutorial](https://github.com/rackt/redux-simple-router/blob/master/README.md#tutorial) looks reasonable enough until you realize that a typical Redux setup will be doing much more than the following two lines:

```js
const createStoreWithMiddleware = applyMiddleware(reduxRouterMiddleware)(createStore)
const store = createStoreWithMiddleware(reducer)
```

Think additional middleware, store enhancers, reducer updating, etc. A [very](https://github.com/rackt/redux/blob/master/examples/real-world/store/configureStore.dev.js) [common](https://github.com/rackt/redux/blob/master/examples/async/store/configureStore.js) [pattern](https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/store/configureStore.dev.js) is to abstract away this setup into a separate function which returns an enhanced `createStore` function, parameterized according to the [Redux spec](http://redux.js.org/docs/api/createStore.html). This is a really clean abstraction.

However, using redux-simple-router@2, `createStore` now requires access to the `history` singleton to pass to `syncHistory`. `createStore` can't initialize it because it needs to be passed around elsewhere (e.g. to react-router), but you can't further parameterize `createStore` without breaking its natural signature (preventing further store enhancement). It also must now hook onto its own store creation so that `listenForReplays` can be called on the store once it's created.

At that point, if you're already entangling all of this history logic in your store creation -- left as boilerplate for the user -- why isn't redux-simple-router just a store enhancer? This would eliminate the need for `listenForReplays` and allow `createStore` to be enhanced by redux-simple-router at the top level. Your history logic stays separate, there's less boilerplace, you don't need to break `createStore`, and you get the same behavior.

timdorr:
react-router embraces singleton histories for version 2.0.0 (coming out soon!): https://github.com/rackt/react-router/blob/master/upgrade-guides/v2.0.0.md#history-singletons-provided

You can similarly define your history in a singleton manner, making passing it around a lot easier and not requiring you to include it as an argument to your `createStore` wrapper.

mjrussell:
@billyjanitsch I agree that this new structure does typically require you pass the history singleton both to your createStore and your Router, but how would a store enhancer in this case improve that? 

As I see it the store enhancer would still need the history access and therefore you just moved the call from inside applyMiddleware to the compose but it still lies in the same code. Unless of course you pass down the store enhancer to this configureStore function but thats not really any better.

Maybe some psuedo js of how it would be used would make it more clear, perhaps Im just missing something.

(just saw @timdorr comment but thought I'd still ask the question)

billyjanitsch:
@timdorr

> react-router embraces singleton histories for version 2.0.0 (coming out soon!)

Ah, thanks for the link, I hadn't seen this! That's a huge improvement, and solves most of my wall of text above aside from `listenForReplays`. :)

@mjrussell 

Sorry, I was unclear. The idea is that your store enhancers should work separately from each other so that you can (conditionally) compose them however you'd like. When `listenForReplays` (which is essentially a user-end store enhancer) depends on some parameter of `applyMiddleware` -- even with access to the `history` singleton -- it breaks this composability.

Perhaps I want to abstract my middleware:
```js
import {createStore, applyMiddleware} from 'redux'
import {syncHistory} from 'redux-simple-router'
import {browserHistory} from 'redux-router' // thanks @timdorr :)
import middleware from './myMiddleware'

const enhanceStore = compose(
  syncHistory(browserHistory),
  applyMiddleware(...middleware)
)

export default enhanceStore(createStore)
```

Or abstract most of my store enhancers:
```js
import createStore from './myCreateStore' // already enhanced by me
import {createHistory} from 'history' // stuck on react-router v1 :(
import myReducer from './myReducer'

const history = createHistory()
const finalCreateStore = syncHistory(history)(createStore)
const store = finalCreateStore(reducer)

render(<App store={store} history={history} />, container)
```

It's not any particular case; it's the general principle of keeping middleware and store enhancers abstractable. I believe that store enhancers were in fact designed to keep middleware from having to do this sort of thing.

mjrussell:
@billyjanitsch Thanks I see where you are going, especially if you split your enhancer into a separate file then you do benefit from some decoupling in that case because you can do the sync to store call inside the enhancer.

taion:
The main constraints here are:

1) The basic middleware is just like any other middleware (other than needing the `history`, which it can get from a singleton exported by some module), and is very simple
2) Without a special history per #138, you need something else after the dev tools to pick up on stuff devtools does
3) Ideally, the store enhancers from (2) and (1) can communicate to each other, to prevent (1) from generating new Redux actions when restoring history location, unless we go with #138 

I'd largely be happy with whatever – the main goal here was just to move to an implementation in (1) that naturally excludes the possibility of updating history twice, through the previous syncing method.

taion:
And I'd add that you'd want to probably completely discard the replay logic when not in development.

tomatau:
@taion, apologies for the delay.

> Can you step through ~https://github.com/rackt/redux-simple-router/blob/master/src/index.js#L84-L102 to see what's happening? Unless you're completely wiping history (in which case we need to push a new action to get the store into the correct state), you should hit the forward branch there, and as long as syncing is set, the history shouldn't update the store.

A POP is being passed to `history.transitionTo` with the correct previous path etc... This seems to just get swallowed. After which, history.listen gets invoked correctly, and the store is returning the correct location.key -- but the path __hasn't__ updated in the browser still.

NB: It's a little bit painful to debug with the DevTools or whatever monitor, repeatedly emitting store updates.

taion:
@tomatau I believe that's https://github.com/rackt/redux-simple-router/issues/164, then? Where do you see the additional store updates come from?

tomatau:
Looks very related to #164 

When I say `store updates` -- I'm talking about from the DevTools or LogMonitor itself, it seems to fire some sort of tick invoking the store.subscribe handler. Only a pain point for triggering the debugger.

pdf:
After this refactor, how are we supposed to access the router state in an immutable store?

mjrussell:
@pdf maybe you are looking for this? - https://github.com/rackt/react-router-redux/pull/218

pdf:
Indeed, thanks @mjrussell

PR #143, opened 2015-12-29
Make coding style more consistent

chentsulin:


jlongster:
We have some big PRs open already that restructure stuff, so this is going to conflict. I'm considering the current code basically frozen because we're about to land a refactoring (see #141).  Thank you very much for the PR (we <3 PRs) but if it's ok with you, I'd rather just focus on the refactor instead of conflicting with it.

chentsulin:
It's great and I will back after refactoring and see if something I can help :+1:

PR #148, opened 2015-12-31
Location of createHistory has changed

casche:
Fixes the README.md QuickStart with the new location of createHistory :smile:


taion:
The current code is right... it's creating a browser history. A bit confusing, maybe. Probably should use the RR 2.x singletons anyway.

jlongster:
According to the [latest docs](https://github.com/rackt/history#installation), the original code is correct. We may need to look more at RR 2 and see if we should change anything.

PR #154, opened 2016-01-02
Add .babelrc file to example/basic

travisperson:
Without the .babelrc, when trying to install and run just the example an error occurs stating:

`Couldn't find preset "es2015" relative to directory "<DIR>/redux-sample-router"`

This is due to babel walking up the directory tree looking for a .babelrc file to use. Once it finds the projects folder it would appear to start looking from that points up again to resolve the presets.

Being that the .babelrc file needs to be in the same directory as the node_modules directory anyways, I've copied the file into the `examples/basic` folder.

The other option is to instead configure babel inside of the package.json itself for the basic example and any other examples down the road.


StevenIseki:
I also have the same issue, would be good to get this in, however the .babelrc in this pull request should be updated to just preset es2015, as stage-2 is not a devDependency for the basic example.

```json
{
    "presets": ["es2015"]
}
```

timjacobi:
I don't think this is the correct solution. The error message is being displayed because babel cannot find the preset in the main project. See https://github.com/rackt/redux-simple-router/blob/f5ef679c4cfcb9018bee64edbfd09f99fff795c7/examples/basic/webpack.config.js#L21

The actual solution to the problem is to correct the readme file to tell the user to run `npm install` in the main project.

StevenIseki:
Yes but if you remove the second part from the webpack.config and follow the advice to:  `// Remove this if running outside of the source.` then you expect it to work just from `npm install` in the **example/basic** project and you should not need to then `npm install` from the main project, so with this in mind adding the **.babelrc** file fixes the issue and the basic app works by itself.

However, as it stands it makes no real sense, because the current version of the basic example is using the latest version of redux-simple-router 2.0 which is not pubishished yet, so the current example will not work unless you use the version from source.

As I mentioned in the following issue, the current version is broken if you try to use it from 1.0.2 https://github.com/rackt/redux-simple-router/issues/157

So you need to use https://github.com/rackt/redux-simple-router/tree/1.0.2/examples/basic

I think this is quite confusing, and the master tag should be the latest stable version and a 2.0 tag should be worked on separately then merged in to master once it is stable, I might not be correct with this though having not worked on such a large project myself. 

I can imagine as this is the official example, it is causing new users quite a few headaches and a barrier to entry using redux with react router.

Back to this commit, it still makes sense to add it, as even in the 1.0.2 tag if you want to run from the npm published version you need to add the .babelrc yourself.

StevenIseki:
I would suggest adding the .babelrc and an update to the README which I added https://github.com/rackt/redux-simple-router/pull/162 will make it a lot easier for new users starting out with redux-simple-router to get the basic example up and running, without also having to `npm install` the source in the main project.

jlongster:
I agree it can be confusing. But the `examples` directory is definitely supposed to always be up-to-date with `master`. If you want to see old examples, you just need to use tags to go back in versions.

The `examples` code is a great way to show people what's coming and how it's actually used in a project. Right now we are in the worst case though, where a new refactoring has landed for a while but it's not quite released yet. I agree that's not ideal but I also don't think it's worth over-worrying about. People expect either either `examples` to be showing the latest `master` code and building from src (extremely helpful for new contributors who want to test changes) or for it to be showing the latest released version. We can't have both, and I think the former has more benefits.

For the latter just use tags. Thanks for your other PR, I'll definitely merge it to explain this better.

> I think this is quite confusing, and the master tag should be the latest stable version and a 2.0 tag should be worked on separately then merged in to master once it is stable, I might not be correct with this though having not worked on such a large project myself.

`master` is where we land work that should be included in the next release. `master` should be *stable*, but it definitely includes *new* work. There isn't much point in keeping `master` basically just the latest released. We have tags for that (there is a tag for every version).

jlongster:
I think this PR is valid though, if there aren't any downsides and this makes it easier to people to run the example without worrying about the main project, that sounds good to me.

timjacobi:
You will still have to run `npm install` in the main package for this to work.

jlongster:
How come? If you use the project from npm you shouldn't need to.

travisperson:
@jlongster currently the coded in this example is not compatible with the version in the package file, this needs to be updated once a release is made.

Till then, you must use the current configuration defined in the webpack to grab the src from the project. If the plan is to always keep the example up to date with the project source and not the current release then things should be reorganized a bit.

timjacobi:
Let me just go over the issues I'm seeing cloning this repo from scratch.
After successfully `npm installing` the example dependencies I try running webpack with `webpack --watch`. This fails because I don't have webpack installed globally (I don't install anything globally, ever). Running webpack with `npm start` though works as npm scripts have access to the locally installed webpack. But as you can see from the log this fails. I then run `npm install` in the main project as the code in `src` depends on these packages and then it works.
```
[tim@localhost Code]$ git clone https://github.com/rackt/redux-simple-router
output truncated
[tim@localhost Code]$ cd redux-simple-router/examples/basic/
[tim@localhost basic]$ npm install
output truncated
[tim@localhost basic]$ webpack --watch
bash: webpack: command not found...
^C
[tim@localhost basic]$ npm start

> rsr-basic-example@0.0.0 start /home/tim/Code/redux-simple-router/examples/basic
> webpack --watch

Hash: d4d836f7b93a92c34c45
Version: webpack 1.12.9
Time: 6278ms
    Asset     Size  Chunks             Chunk Names
bundle.js  1.02 MB       0  [emitted]  main
    + 382 hidden modules

ERROR in /home/tim/Code/redux-simple-router/src/index.js
Module build failed: Error: Couldn't find preset "es2015" relative to directory "/home/tim/Code/redux-simple-router"
output truncated
^C[tim@localhost basic]$ cd ../..
[tim@localhost redux-simple-router]$ npm install
output truncated
[tim@localhost redux-simple-router]$ cd -
/home/tim/Code/redux-simple-router/examples/basic
[tim@localhost basic]$ npm start

> rsr-basic-example@0.0.0 start /home/tim/Code/redux-simple-router/examples/basic
> webpack --watch

Hash: 47dcf92d6dd6ba53be33
Version: webpack 1.12.9
Time: 11077ms
    Asset     Size  Chunks             Chunk Names
bundle.js  1.03 MB       0  [emitted]  main
    + 382 hidden modules

```

jlongster:
> @jlongster currently the coded in this example is not compatible with the version in the package file, this needs to be updated once a release is made.
> 
> Till then, you must use the current configuration defined in the webpack to grab the src from the project. If the plan is to always keep the example up to date with the project source and not the current release then things should be reorganized a bit.

You have conflicting statements. Yes, you should be using the webpack config to grab the current src because that's what it needs to work. `examples` on `master` should work with `master`, not the latest version. If you want to see the latest version you need to get it from the version tag.

So yes, the plan is to keep it up to date with the current source, which seems like your saying works with the current config.

I think I see what you may be referring to: the example app has a dependency on `redux-simple-router` in the package.json. We probably should just remove that.

jlongster:
@timjacobi I don't install it globally either. I have `node_modules/.bin` in my path which lets you use local binaries. Either way, this has nothing to do with redux-simple-router.

This PR was supposed to fix that babel error. Anyone know why it does not?

timjacobi:
@jlongster the last bit of my report describes it. The code in `src` has dependencies that need to be installed by running `npm install` in the main project.

jlongster:
Oh, if you did not remove the webpack alias and are pulling it from source, then yes you need to run `npm install` from the main project. Sorry if that's not documented well, I will make sure it is now.

StevenIseki:
Thanks for clearing all of that up @jlongster 

Just a quick note, now with your update to the [README](https://github.com/rackt/redux-simple-router/commit/6760e40baf5f4ee65d2de3585f46a02493eaba06#diff-f3ae7e76f103bdc637ff137af4d9abfe) you can probably remove my part about removing the alias, as you mention this in step 2. So you can remove line 13 and 14

I agree with you it's a good idea to have the basic example up to date with the current source, only thing is it's annoying that it breaks the example. So even if you do install from the published version e.g. 1.0.2, the example will not work as `syncReduxAndRouter` has changed to `syncHistory`. So you then would have to jump back to the tag 1.0.2 version which is not too bad I guess...

hmm... my personal opinion is that you should forget trying to make it work for both, and just make basic example work with the latest source. In the basic example README refer users to the list of stable examples and mention that basic example uses the latest source and is for development, it does not use the most recent published version of redux-simple-router.

Some nice examples are coming in at the moment, so this might make more sense, I just set up one based on the basic example 1.0.2 with a few small additions, css-modules, isomorphic-fetch and express [redux-simple-router-example](https://github.com/StevenIseki/redux-simple-router-example)

Anyway this is an awesome library, thanks heaps for it :) I'm just worried for new users who want to use redux, react and react router, as they will come straight to redux-simple-router and will want some quick example which run first time from the stable version.

travisperson:
> I think I see what you may be referring to: the example app has a dependency on `redux-simple-router` in the package.json. We probably should just remove that.

That is exactly what I'm referring too. The webpack configuration also states you can comment out part of it to use the version defined in the package file. Which should of course be removed too as it will no longer be true.

In my previous comment I was simply stating that there were two possible actions. Keep the example as a latest to the current source, or make it an example of the latest published version.

Id prefer it always be up to date with the source, that way users simply can check outthe tag of the published version they are using to see a working example. Which appears to also be your intentions.

timjacobi:
I like the idea of having the example working with the latest source but I find it very confusing that it is solved with webpack. Also when you move the example somewhere else it will break as the reference to the source becomes invalid.
Why not just add this to `package.json`
```JSON
"dependencies": {
    "redux-simple-router": "git+https://github.com/rackt/redux-simple-router.git"
  }
```

StevenIseki:
I imagine `webpack --watch` is so as the developers are making changes to the source and testing in the basic example it is watching for changes. In my opinion webpack is an awesome tool and very common in react and redux applications, so it doesn't make much sense to update the dependency to be the git repo.

As I mentioned as the basic example uses the latest source and is really only geared towards developers working on the source. It should not be referred to as an example for new users, and they should be redirected to the examples in the README. 

@timjacobi if you have some time, it might be an idea for you to make a copy of the basic example with as few dependencies as possible, using v1.0.2 and make a PR to add the link to the README examples. I added a similar example myself, but it had a few additions css modules and isomorphic fetch so might not be as simple for new users, although it was very similar to the basic example.

timjacobi:
Ah I see. I totally didn't get that it was an example for development of the library only. That was a bit misleading for me. I'm happy to provide a self-contained example if there is interest for one. I personally think its best for new users to have that so they can try out the library before deciding to use it.

PR #137, opened 2015-12-24
feat(devtools): update to devtools stable

justingreenberg:
note: react is a peerDep of devtools in v3, so we need to include as devDep for now

PR #167, opened 2016-01-06
update (lib and example) to devtools stable

### LINKED ISSUE #170: Worth updating the examples?
At least the basic example is out of date, pretty much in conflict with the README, which is confusing.

I'd be happy to update if we think the examples are going to stick around. A lot of activity on this project, so I don't want to update something that won't be around in a month or so :)

justingreenberg:
@jlongster here's a fresh PR for devtools, rebasing #137 was too painful... i can submit separate PR updating example to es modules and rackt style if you want, let me know :)

resolves #170


jlongster:
Awesome, thanks!

PR #177, opened 2016-01-08
[fix] examples/basic

langpavel:
Small fix and refactoring, see diff.


langpavel:
Travis test failed for node v4, but not because of examples (there is no test)
May be broken master or test env?

justingreenberg:
these changes were included in #167

langpavel:
@justingreenberg Oh, thanks. I close this.

PR #178, opened 2016-01-09
Do not swallow return value of next() in middleware

langpavel:
Correct way is to pass unknown actions to `next()` and `return` it's value.

Current implementation can break other middlewares


jlongster:
You're totally right! This is the kind of stuff I wanted to find before releasing 2.0.0. Thanks so much!

PR #180, opened 2016-01-09
Prepare README for 2.0

Furizaa:
This PR updates the readme.md to reflect the API changes introduced mostly in https://github.com/rackt/redux-simple-router/pull/141 by @taion.

It also adds a little section for server-side usage - explaining that the memoryHistory has to be hydrated with the current location from the server for redux-simple-router to determine the correct initialState.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,20 +47,26 @@ Let's take a look at a simple example.
 ```js
 import React from 'react'
 import ReactDOM from 'react-dom'
-import { createStore, combineReducers } from 'redux'
+import { compose, createStore, combineReducers, applyMiddleware } from 'redux'
 import { Provider } from 'react-redux'
 import { Router, Route } from 'react-router'
 import { createHistory } from 'history'
-import { syncReduxAndRouter, routeReducer } from 'redux-simple-router'
+import { syncHistory, routeReducer } from 'redux-simple-router'
 import reducers from '<project-path>/reducers'
 
 const reducer = combineReducers(Object.assign({}, reducers, {
   routing: routeReducer
 }))
-const store = createStore(reducer)
 const history = createHistory()
 
-syncReduxAndRouter(history, store)
+// Sync dispatched route actions to the history
+const reduxRouterMiddleware = syncHistory(history)
+const createStoreWithMiddleware = applyMiddleware(reduxRouterMiddleware)(createStore)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,29 +81,42 @@ ReactDOM.render(
...
 
+### Server-side usage
+
+Using redux-simple-router on the server works nearly the same as using it on the client-side - except for the creation of the history. To make that work correctly, you have to hydrate the history with the current location passed from the server. This is necessary for redux-simple-router to determine the correct initial state.
+
+```js
+import { createHistory } from 'history'
+
+// Creates a memoryHistory with a necessary initial pathname.  
+const history = createHistory({ entries: [ req.url ] });
```
jlongster: Can you remove this section? I don't think this is the right way to do server-rendering. Why do you have a history object at all? You should be using the lower-level `match` API directly.

jlongster:
Thanks a lot for this! I'll merge one you remove that section. That's not the right place for an implementation detail; for server-rendering I'd rather link to proper example once we get there.

Furizaa:
Removed, rebased, squashed :)

Server rendering works pretty neat for me once I figured out that I have to hydrate a memoryHistory and pass that to `match` for redux-simple-router to determine a non-empty initial state. But you are right - perhaps I'm able to create a failing test for this in the following days.

jlongster:
Personally I update the state after creating the server-side store with a `routeActions.push` action. We should have better docs about all this but let's get 2.0.0 out first. Thanks!

ericdfields:
@jlongster routeActions is undefined in the current npm package. happy to file a separate issue…

jlongster:
Nope, I'm about to release 2.0.0 so we updated the readme in preparation for it.

ericdfields:
I sorta figured that out as the day went on :). Using 2.0 on dev right now in anticipation of the release, so I should be all ready.

georgesboris:
Is the 2.0.0. release scheduled for today?
The lack of routeActions on the npm release was a really unexpected hold up on my development.
Should I wait for 2.0.0. (as in a few hours) or must I use an older version?

(I know this question would be better off in a separate issue, but since it has been brought up here and @jlongster said it should not be dealt with since 2.0.0 is about to be released, i figured i'd should just post it here)

jlongster:
Yes it will be released today.

Issue #189, opened 2016-01-13
syncHistoryToStore makes createHistory throw an error 


georgesboris:
Just tried to upgrade to the current source as the npm one is not working.
I've reset my configurations based on the new docs. But now I'm getting this error:

```
createHistory.js?cf40:39
Uncaught TypeError: Cannot read property 'pathname' of undefined
```
this error is being thrown here:
```
function locationsAreEqual(a, b) {
  return a.pathname === b.pathname && a.search === b.search &&
  //a.action === b.action && // Different action !== location change.
  a.key === b.key && _deepEqual2['default'](a.state, b.state);
}
```

but if I omit the `reduxRouterMiddleware.syncHistoryToStore(store);`
no error is thrown (but the router does not work).

timdorr:
`syncHistoryToStore` is only needed if you're using Redux Dev Tools. If not, you can omit it. We should make the docs more clear about this, though.

jlongster:
`locationsAreEqual` does not exist in the latest code, so you probably don't have things installed correctly.

@timdorr Not a bad point. We can add it somewhere, but people should still call it so whenever they try the devtools it "just works".

jlongster:
I'm releasing 2.0.0 right now, and this code does not exist in there.

jlongster:
@timdorr In fact, I think I'm just going to rename that method. I didn't realize that it was 100% just for replays. I'm going to rename it to `listenForReplays`. I'd like to go ahead and push 2.0.1 out though.... Anyone happen to see this and have thoughts on that?

jlongster:
https://github.com/rackt/redux-simple-router/pull/191

PR #191, opened 2016-01-14
rename syncHistoryToStore to listenForReplays

jlongster:
@timdorr mentioned in https://github.com/rackt/redux-simple-router/issues/189#issuecomment-171524092 that this function is purely for devtools. I sort of forgot that it is completely for devtools, and because if this, I think we should rename the function. I propose `listenForReplays`, given that that's the usual use case. It might not be a replay that triggers it, but that's not a huge deal, it's something *like* a replay.

We can also mark this optional; people that don't use the devtools don't need it.

### DIFF
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -31,7 +31,7 @@ const finalCreateStore = compose(
   DevTools.instrument()
 )(createStore);
 const store = finalCreateStore(reducer);
-middleware.syncHistoryToStore(store);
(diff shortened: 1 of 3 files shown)


jlongster:
I don't feel good pushing this out without any review, so I'm going to wait to push out 2.0.1 until tomorrow morning.

timdorr:
LGTM.

Issue #196, opened 2016-01-15
It appears you have provided a deprecated history object to `<Router/>`


gabrielenosso:
Using the code in the tutorial as it is results in receiving a Warning by react-router:

It appears you have provided a deprecated history object to `<Router/>`, please use a history provided by React Router with `import { browserHistory } from 'react-router'` or `import { hashHistory } from 'react-router'`. If you are using a custom history please create it with `useRouterHistory`, see http://tiny.cc/router-usinghistory for details.

As you can see from this link, https://github.com/rackt/react-router/blob/master/upgrade-guides/v2.0.0.md#using-custom-histories, the creation of the history should change in:

import { useRouterHistory } from 'react-router';
const history = useRouterHistory(createHistory)({ queryKey: false });

timdorr:
You're right! Would you mind submitting a PR to fix this?

PR #198, opened 2016-01-15
Update docs to reflect internal history in react-router 2.0

wesbos:
Just fixed an issue in my app where I was using react-routers internal history and then another one via this tutorial.


timdorr:
Fixes #196 
Thanks!

PR #200, opened 2016-01-15
Removed old line from doc example

khlieng:


timdorr:
Ah, missed this. Thanks!

PR #201, opened 2016-01-15
replace react-router's BrowserHistory with history

justingreenberg:
@timdorr #198 replaced `history.createHistory` with the `browserHistory` module from `react-router`

i'm a little confused... what is the recommended usage? should we be using `react-router`'s internal history or handing it a factory? my understanding was that `redux-simple-router` is supposed to integrate with history via react-router's useRouterHistory enhancer—i have updated the example to reflect this. i used a similar [implementation for another migration](https://github.com/davezuko/react-redux-starter-kit/pull/458/files).

any guidance is appreciated :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,10 @@ Let's take a look at a simple example.
 ```js
 import React from 'react'
 import ReactDOM from 'react-dom'
-import { compose, createStore, combineReducers, applyMiddleware } from 'redux'
+import { createStore, combineReducers, applyMiddleware } from 'redux'
 import { Provider } from 'react-redux'
-import { Router, Route, browserHistory } from 'react-router'
+import { createHistory } from 'history'
+import { Router, Route, useRouterHistory } from 'react-router'
 import { syncHistory, routeReducer } from 'redux-simple-router'
 import reducers from '<project-path>/reducers'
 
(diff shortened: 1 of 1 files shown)


mjrussell:
@justingreenberg you can use any history you want as long as it conforms to the specification. `React-Router`'s internal history is just a wrapping around history from `history`. They also provide some handy singleton reference. The  #198 PR did help clear up some confusion where when you go to split up your React-Router render and your createStoreConfig, some people were accidently importing createHistory twice (instead of passing down a reference to single history) and having them not be in sync. By using React-Router's provided history you dont have this issue because they give you a singleton. 

I think the docs might be better improved by showing both cases so people don't think there is a hard requirement of one vs the other.

justingreenberg:
@mjrussell thanks for the explanation, that was my understanding as well. 

there are some limitations associated with using react-router's browserHistory, so i'm just not sure it should be the canonical example. as you point out—react-router's internal history returns a singleton, so for example, working with enhancers like useBasename is not possible

passing history.createHistory to react-router.useRouterHistory [returns a new factory composed with useQueries and useBasename](https://github.com/rackt/react-router/blob/master/modules/useRouterHistory.js#L6) that you can invoke with an options object to create the singleton, ie:

```js
import { useRouterHistory } from 'react-router'

const history = useRouterHistory(createHistory)({
  basename: '/app',
  queryKey: false,
  parseQueryString: (s) => s // etc
})
```

this seems like a more explicit, durable example imo. either way, the enhancer use case should probably be documented, maybe added as a recipe or something

timdorr:
The `browserHistory` singleton is there as a convenience. If it's not sufficient, you can certainly create your own. But learning that API is best left to react-router and not this project.

We're assuming you've already gotten started with router and now want to hook it up to redux. The example should be minimalistic with outside libraries and emphasize as much of our own API as is possible.

PR #208, opened 2016-01-17
Make UPDATE_LOCATION follow FSA

### LINKED ISSUE #206: Back FSA actions
I'm curious why you decide to move from FSA in 2.x.x

SimenB:
Fix #206

Could use `redux-actions` to create the action creator, but that seems overkill

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,7 @@ const SELECT_STATE = state => state.routing
 function transition(method) {
   return arg => ({
     type: TRANSITION,
-    method, arg
+    payload: { method, arg }
   })
 }
 
@@ -23,7 +23,7 @@ export const routeActions = {
 function updateLocation(location) {
   return {
     type: UPDATE_LOCATION,
-    location
+    payload: location
   }
 }
 
@@ -33,7 +33,7 @@ const initialState = {
   location: undefined
 }
 
-export function routeReducer(state = initialState, { type, location }) {
+export function routeReducer(state = initialState, { type, payload: location }) {
   if (type !== UPDATE_LOCATION) {
     return state
   }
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks! This is a good start. Can you add the TRANSISTION actions as well?

SimenB:
Ah, didn't bother with it as I thought it was internal.

Use `redux-actions`, or still do it manually?

timdorr:
Spellchecker fail: TRANSITION actions.

timdorr:
Manually. I don't think we need to add another dependency.

SimenB:
@timdorr Done.

timdorr:
Awesome! Thank you!

SimenB:
@timdorr I didn't realize 2.0 was out of beta.

This is a breaking change, isn't it?

timdorr:
Shouldn't really be. The API is the same. Action shapes are internal and
subject to change within the same minor version.

SimenB:
aight :smile:

jlongster:
Well, not sure they are really internal, since people can listen to the `UPDATE_LOCATION` action. This is the sort of thing I hate about semver though, I don't want to release a whole 3.0.0 version just for this, especially since it's a quick change right after 2.0.0.

I don't really know how often people listen to that action, probably not much. We should definitely document it in the changelog though.

SimenB:
Want me to add a PR for the changelog?

valscion:
> Well, not sure they are really internal, since people can listen to the `UPDATE_LOCATION` action.

Yeah, we do that. Glad I checked on the changelog before upgrading from beta version as this would've broke our stuff :sweat_smile: 

I agree that semver sucks in these cases. But I also don't really care at all even if this library's version would change from 2.0 to 3.0 within 4 days. It only tips me off to read the changelog if the major version changes — if there's only one change, it shouldn't be a hassle to upgrade immediately :smile:

I'm just glad I didn't get burned before reading this issue ¯\\\_(ツ)_/¯

PR #213, opened 2016-01-18
fix example by updating to 2.0.3

barrystaes:
Fixes fatal runtime error `Uncaught TypeError: middleware.listenForReplays is not a function`.

### DIFF
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -8,7 +8,7 @@
     "react-redux": "^4.0.0",
     "react-router": "^1.0.0",
     "redux": "^3.0.4",
-    "redux-simple-router": "2.0.3"
+    "redux-simple-router": "^2.0.3"
   },
   "devDependencies": {
     "babel-core": "^6.1.21",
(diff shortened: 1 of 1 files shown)


timdorr:
That shouldn't matter, as the example aliases to whatever you have checked out above it in git. Check the webpack config.

barrystaes:
I had an issue when using release 2.0.3 so i wanted to compare it to the
example at the 2.0.3 tag in the repo. This is where i encountered the issue.

If its relevant, i only `npm install`ed the example not the main
redux-simple-router package.
Op 18 jan. 2016 15:06 schreef "Tim Dorr" <notifications@github.com>:

timdorr:
Can you make it a caret range then? That should future-proof us against any updates until a major version, in which case there will probably be an updated example anyways.

barrystaes:
Good idea, done.

Ideally the example gets tested prior to a release. I see the resolve code would devs to quickly use their latest version. Nice!

timdorr:
Ideally we could test our examples, [just like Redux does](https://github.com/rackt/redux/blob/master/examples/testAll.js). It's part of the test suite over there. I'd be really jazzed for someone to add those at some point :+1:

PR #218, opened 2016-01-20
Support ImmutableJS: use getLocationState instead of getRouterState as mentioned in #140

stepankuzmin:
This works for me with custom reducer:
```js
import { UPDATE_LOCATION } from 'redux-simple-router'

const initialState = new Immutable.Map({
  location: undefined
})

export default function routeReducer(state = initialState, { type, payload: location }) {
  if (type !== UPDATE_LOCATION) {
    return state
  }

  return state.merge({ location })
}
```

and this `selectLocationState`

```js
  const selectLocationState = function (state) {
    const location = state.get('routing').get('location')
    return location ? location.toJS() : location
  }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,11 +126,11 @@ _Have an example to add? Send us a PR!_
 
 Call this to create a middleware that can be applied with Redux's `applyMiddleware` to allow actions to call history methods. The middleware will look for route actions created by `push`, `replace`, etc. and applies them to the history.
 
-#### `ReduxMiddleware.listenForReplays(store: ReduxStore, selectRouterState?: function)`
+#### `ReduxMiddleware.listenForReplays(store: ReduxStore, selectLocationState?: function)`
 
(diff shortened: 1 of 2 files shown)


timdorr:
This is really less about supporting Immutable explicitly, and more about being less prescriptive about the return shape of our `selectRouterState`/`selectLocationState`. That seems like a positive API change to me.

timdorr:
I'm going to leave this open until we want to do a release, since it affects the docs on master. I don't want to confuse people using the published version when the behavior differs from the docs. 

In the meantime, you should be able to do this with your `selectRouterState`:
```js
const selectLocationState = function (state) {
    const location = state.get('routing').get('location')
    return { location: location ? location.toJS() : location }
  }
```

stepankuzmin:
Yep, thanks for reply @timdorr! I've also updated `README.md` in this PR to track changes

PR #241, opened 2016-01-26
Update README with info on react-redux route params

robertfall:
Add an example of getting route params from react-router in mapStateToProps.


timdorr:
Thanks!

PR #253, opened 2016-02-02
Update usage example in README

zaynetro:
Now you need to manually include `history` lib for creating history object. See [https://github.com/rackt/react-router/issues/1455#issuecomment-138295614](https://github.com/rackt/react-router/issues/1455#issuecomment-138295614) for more details


zaynetro:
It works with react-router v2 release candidate. Failed to install correct versions of all dependencies.

Issue #257, opened 2016-02-03
Clarifying the goals of this project


gaearon:
I think I’m getting a little confused by the goals and the scope of this project. I’d appreciate @jlongster weighing in on this.

I see a few things it does right now:

* **(1) Provides `state.routing` for other middleware, components, and action creators to read.** I believe this is problematic: it doesn’t match the rendered route because of the potential for async transitions in between. This is the reason we warn people not to use `state.routing` in `mapStateToProps()` which I believe is confusing, and even if they don’t use it in components, they still might use it in action creators and expect it to correspond to the router’s current state. So I’m not sure whether this is a good feature to keep.

* **(2) Provides an actions-first interface to side effects of `history`.** While this may be convenient, I think the reason people want this is not because actions are the best interface for causing routing changes, but rather because people don’t want to deal with `context` or pass `history` around. Many people are also feeling pressured to express everything with Redux primitives (actions) even if the problem at hand is not expressed with them well. Consider that `history.pushState()` is not much different from `fetch()` (it’s a side effect) but nobody expects us to provide “`fetch` action creators”. The real problem here seems to be the perceived difficulty of passing `history` around or importing it, rather than the need for history updates to be actions.

* **(3) Allows replaying the history changes from DevTools or another action replay mechanism.** This is where it gets interesting. It seems to me that the ability to persist routing state together with everything else and re-render the app based on current state rather than what the browser thinks was the original point of this library. This feature seems to explain the need for action creators: you want navigations to be in the log so they can potentially be replayed. However such functionality would only be useful in dev (or otherwise special) environment.

As we evolve (or not evolve) this library, I want to better understand which of these points are actually important to the author, and which are just a byproduct of current user habits and their desire to express everything “the Redux way”.

Essentially, I want to understand what value `react-router-redux` brings over `react-router` + `history` in production mode other than the familiar mental model (“just dispatch actions”), the ability to reach out into `state.routing` (which can be dangerous and may not reflect the UI but can be convenient for apps without async transitions), and not having to think about how to access `history` without passing it around.

In no way am I trying to diminish the work being done here. I just want to figure out for myself whether presistence/time-travely part of this library could be a separate dev-only project, and whether what would be left is not essentially a feature request for better documentation and less ceremony when accessing `history` from action creators and other middleware.

Thank you!

timdorr:
[...]

Issue #207, opened 2016-01-17
Dev Tools reset wipes out location state


srph:
I'm not sure if this is a bug or an issue with my code. Let me know if I'm submitting this to the wrong place.

Whether this was intended or not, please let me know.

## Behavior 
The state.routing.location key loses its value when the state is reset.

## Demo
https://github.com/chirigarikku/redux-simple-router-2-reset-state-bug

![preview](https://github.com/chirigarikku/redux-simple-router-2-reset-state-bug/raw/master/preview.gif)

jasonh42:
Getting this issue as well. Using Redux DevTools, if you load a page, click "Reset", go between between pages, then try to replay the route changes, you get:

Uncaught TypeError: Cannot read property 'action' of undefined

If you don't Reset the state, then no issue occurs during replay. I wasn't getting this issue in 1.x.

timdorr:
Looks like we need to reapply something like #6. It should [go here](https://github.com/gaearon/redux-devtools/blob/72cd8d050509b416fb3d65531978f5bd301c0cff/src/instrument.js#L168-L178). Anyone want to try their hand at a PR?

srph:
i'll attempt

srph:
looks like we'll need to parse the queries as well. 

what i got so far:

```es6
var initialState = typeof window === undefined ? {} : {
  location: {
    pathname: location.hash ? location.hash.substr(1) : '',
    search: (location.search ? ('?' + location.search) : ''),
    hash: '',
    action: '',
    key: '',
    query: qs.parse(location.query.replace(/\+/g, '%20'))
  }
};
```

srph:
Can anyone point out if the <strike>solution</strike> hack I've thought of is valid?

jlongster:
This is inherit in how this works -- we need to read the location first from `history`. What is actually throwing the error, is it something inside this library? I suppose user code is running with this default state. Maybe we should set the initial location to a blank object like `{ pathname: '' }` instead?

vga-:
I couldn't get it to work by setting the initial state to a blank object.

Following @timdorr's suggestion, I've set the initial state to the window's location, string, and hash as follows:

```javascript
const initialState = typeof window === undefined ? {} : {
  location: {
    pathname: window.location.pathname,
    search: window.location.search,
    hash: window.location.hash
  }
};
```

[...]

PR #259, opened 2016-02-05
Split history syncing from action creators

### LINKED ISSUE #193: First mutation
I'm using the latest version of this repository.
The `@@router/UPDATE_LOCATION` action do not reach the end of my middleware chain during the first mutation of the state.
I'm wondering if this is intentional, or an issue with this project (it's working with `redux-router`).

Here is my middleware chain in case you ask:
```js
import promiseMiddleware from 'redux-promise';
import thunk from 'redux-thunk';

...

  middleware = applyMiddleware(
    promiseMiddleware,
    thunk,
    reduxRouterMiddleware,
    analyticsMiddleware
  );
```

The issue for me is that, I want  to send every `@@router/UPDATE_LOCATION` event to google analytics but the first page is missing.

### LINKED ISSUE #252: Restore router state

timdorr:
#### Big thanks to @gaearon for opening up this can of worms and writing a bunch of code to fix it!

Here's the history (pun intended) of what led to this: 
**https://github.com/rackt/react-router-redux/issues/252#issuecomment-178980683 ->  #255 -> #257 -> rackt/redux#1362 (PR)**

A big giant chunk of this code is copypasta from Dan's work on that Redux PR. I just mashed some buttons until it looked right.
## Changes

The provided action creators and middleware are now separate from the history<->state syncing function. For the vast majority of cases, using action creators to trigger navigation is obsoleted by React Router's [new history singletons](https://github.com/rackt/react-router/blob/master/upgrade-guides/v2.0.0.md#history-singletons-provided) provided in 2.0. Building this functionality in by default and coupling it to our history syncing logic no longer makes sense. 

We now sync by enhancing the history instance to listen for navigation events and dispatch those into the store. The enhanced history has its `listen` method overridden to respond to store changes, rather than directly to navigation events. When this history is provided to `<Router>`, the router will listen to it and receive these store changes. This means if we time travel with the store, the router will receive those store changes and update based on the location in the store, instead of what the browser says. Normal navigation events (hitting your browser back/forward buttons, telling a history singleton to `push` a location) flow through the history's listener like normal, so all the usual stuff works A-OK.
## TODOs
- [x] Add tests
- [x] Update the documentation
- [x] Find bugs/edge cases and fix them as best we can.
- [x] Fix server side rendering

Fixes #193
Fixes #252


timdorr:
This should pass tests now. Namely because there's only one left :smile:

src/sync.js
```diff
@@ -0,0 +1,128 @@
...
+    const handleStoreChange = () => {
+      const locationInStore = getLocationInStore(true)
+      if (currentLocation === locationInStore) {
+        return
+      }
+
+      // Update address bar to reflect store state
+      isTimeTraveling = true
+      currentLocation = locationInStore
+      history.transitionTo(Object.assign({},
```
gaearon: Let's use spread here so we don't have to worry about polyfilling Object.assign(). Babel still doesn't replace this call automatically does it?

src/sync.js
```diff
@@ -0,0 +1,128 @@
...
+    // Tell the store to update by dispatching an action
+    store.dispatch({
+      type: LOCATION_CHANGE,
+      locationBeforeTransitions: location
+    })
+  }
+  unsubscribeFromHistory = history.listen(handleLocationChange)
+
+  // The enhanced history uses store as source of truth
+  return Object.assign({}, history, {
```
gaearon: Spread here as well

src/sync.js
```diff
@@ -0,0 +1,128 @@
...
+    listen(listener) {
+      // History listeners expect a synchronous call
+      listener(getLocationInStore(true))
+
+      // Keep track of whether we unsubscribed, as Redux store
+      // only applies changes in subscriptions on next dispatch
+      let unsubscribed = false
+      const unsubscribeFromStore = store.subscribe(() => {
+        if (!unsubscribed) {
+          listener(getLocationInStore(true))
```
gaearon: We need to add conditions here that only call the listener it the location field has changed. We will probably need to keep lastPublishedLocation in the listen() closure and bail out when they are equal so other store changes do not touch the router.

src/sync.js
```diff
@@ -0,0 +1,128 @@
...
+
+      // Let user unsubscribe later
+      return () => {
+        unsubscribed = true
+        unsubscribeFromStore()
+      }
+    },
+
+    // It also provides a way to destroy internal listeners
+    unsubscribe() {
```
gaearon: How does disposing history enhancers compose? Is this the only history enhancer with internal resources?
If normally enhancers free resources when last subscription dies, can we do this here? If not, why?
If we decide unsubscribe() is way to go, should we not call it on underlying history too, if it exists, to ensure this enhancer is composable with other possible enhancers that may also have internal resources?

In the original PR I called this dispose() to be consistent with disposables in Rx and make it clear that this method is not related to listen() subscriptions. And that using enhanced history() after dispose() won't work. Essentially we might want to throw it you attempt to listen() to a disposed history to prevent silly errors. What do you think?

gaearon:
Thank you very much for getting this moving! I might be able to find some time to contribute in the next week but this is already shaping up nicely 🙃

src/sync.js
timdorr: It does. And Travis' version of Firefox (31) is old enough not to have Object.assign built in. So it doesn't really matter either way, but I will probably switch it to a spread. Haven't made it this far down in the file yet :smile:

gaearon: Does it need some special flag?
http://babeljs.io/repl/#?experimental=false&evaluate=false&loose=false&spec=false&code=Object.assign(a%2C%20b)%0A%0A

SimenB: I think you need [this](https://www.npmjs.com/package/babel-plugin-transform-object-assign)

ryanflorence:
:clap:

webmasterkai:
Can someone confirm for me that this is just syncing state with rackt/history and the use of react-router is secondary? Any objections to me making a pull request against the synchronicity with the change suggestions above?

gaearon:
>Can someone confirm for me that this is just syncing state with rackt/history and the use of react-router is secondary?

This is correct. `react-router` in the library name is just for marketing purposes because that’s what people google for.

>Any objections to me making a pull request against the synchronicity with the change suggestions above?

I’m pretty sure any contributions are welcome!

webmasterkai:
Was it decided somewhere how defaultState based on current location should be set? This is something that was handled before that seems to be missing now. Having to manually do the following seems a bit inelegant.

```javascript
const history = createHistory()
let locationBeforeTransitions
history.listen(loc => locationBeforeTransitions = loc)()
const initialState = {
    routing: {
    locationBeforeTransitions
  }
}
const store = createStore(reducer, initialState)
```
Is it the job for a store enhancer?

webmasterkai:
Just a note that the discussion at https://github.com/rackt/redux/pull/1362 is probably the best place to start pull documentation updates from.

I understand it is discouraged to read location information from the store if a user is using react-router. But with this library I see little reason to continue using react-router. The rest of my components read state and render accordingly. Why should `location.pathname` be any different? Why can't I parse location within a container and render components accordingly?

timdorr:
Router let's you do lots of neat stuff. At its core, you can declaratively define your application structure by URL. You can also do things like adding transition hooks or loading parts asynchronously. You're free to do that on your own, but it's a lot more work that way.

webmasterkai:
I understand there is a lot of power with Router. I have used it since its inception and survived every api change. I think what I'm realizing is that location contains a minuscule amount of state when compared to an entire application. Yes it's probably the slice that dictates the biggest changes but `Base64.encode(JSON.stringify(getState())).length > (location.basepath.length * 100)`. Anyway, you are all my heroes. I love you all and all your codes are amazing. Thanks for the merge. I'll try to help more tomorrow!

gaearon:
>Yes, it is set with an action on invocation of syncHistoryWithStore(). However, dispatching an action afterward feels like a hack when it's possible to set initialState correctly.

Can you please explain the use case for having location in the store prior to the first syncHistoryWithStore() call? I don’t understand how you would use it and why.

webmasterkai:
> Can you please explain the use case for having location in the store prior to the first syncHistoryWithStore() call? I don’t understand how you would use it and why.

Analytics middleware that sends actions to a server. That middleware will need special logic to understand if it's the first time the location is being set or if it is due to another reason. Entry point and navigation points are recorded differently.

API middleware that automatically requests different data depending upon if it's the initial load or a navigation change.

On a fundamental/semantic level, the action type is `LOCATION_CHANGE`. The location is not "changing", it is just a reflection of what the `initialState` should have been.

webmasterkai:
Currently `syncHistoryWithStore` updates the store based on history changes always and updates history based on store changes optionally.

history -> store (always)
store -> history (optional)

In thinking about documentation, I'd like to request that the `adjustUrlOnReplay` option be renamed to `syncStoreToHistory` instead. Thoughts?

gaearon:
@webmasterkai 

Nothing prevents us from putting `isInitialLocation: true` in the action if that is what you need.

timdorr:
> Analytics middleware that sends actions to a server. That middleware will need special logic to understand if it's the first time the location is being set or if it is due to another reason. 

That should listen to the history for updates, rather than be a redux middleware. And you can add a check to ensure the first time it runs it takes some special code path. Same thing applies for other examples. You can basically define your own initialLocation [the same way we do in the current version of this library](https://github.com/rackt/react-router-redux/blob/master/src/index.js#L50). Just do that before you do any React.render'ing.

gaearon:
In any case using middleware for analytics in the way it was previously designed was wrong. You would miss all POP actions from the browser Back button because it would not have been an action.

The current approach is more correct, and determining the first navigation is as easy as keeping a local variable. But definitely we can add a flag to the action itself if it helps.

webmasterkai:
@gaearon @timdorr I've realized my real issue is that the only way to get the current location from history is via listen, set variable, unsubscribe.

@timdorr Analytics middleware must be middleware because its purpose is to listen for actions and send a copy of them to a server.

@gaearon I haven't yet been about to use react-router-redux but I'm looking forward to now.

timdorr:
> Analytics middleware must be middleware because its purpose is to listen for actions and send a copy of them to a server.

You're free to use the action creators and middleware that will still be provided in this version of the library. Those actions can be listened to by middleware. The initial location should still be obtained directly from your history, however.

README.md
```diff
@@ -44,25 +44,25 @@ import ReactDOM from 'react-dom'
...
+  routing: routerReducer
 }))
 
-// Sync dispatched route actions to the history
-const reduxRouterMiddleware = syncHistory(browserHistory)
-const createStoreWithMiddleware = applyMiddleware(reduxRouterMiddleware)(createStore)
+const store = createStore(reducer)
 
-const store = createStoreWithMiddleware(reducer)
+// Sync dispatched route actions to the history
```
gaearon: TODO:

I think this comment should be backwards. We are not syncing "route actions to history". We are syncing history changes to the store. Or "using store as source of truth for the history". Or "routing history changes through the store".

Saying that we "sync dispatched actions to history" is a misleading leftover comment from how it used to work.

timdorr: The whole thing needs to be reworked. I'll likely just scrap it and restart from scratch.

timdorr:
Got all the low hanging fruit taken care of on the test suite. Also reorganized it a bit, since there are a number of components that don't need browser testing anymore. Still needs proper testing on the syncing functions.

README.md
```diff
@@ -73,7 +73,7 @@ ReactDOM.render(
 )
 ```
 
-Now you can read from `state.routing.location.pathname` to get the URL. It's far more likely that you want to change the URL more often, however. You can use the `push` action creator that we provide:
+Now you can read from `state.routing.locationBeforeTransitions.pathname` to get the URL. It's far more likely that you want to change the URL more often, however. You can use the `push` action creator that we provide:
```
gaearon: Here we should warn why it's called so. In particular that it may not reflect the current UI state, and that we don't recommend reading it from the UI of the user also uses async transitions.

src/reducer.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * This action type will be dispatched when your history
+ * receives a location change.
+ */
+export const LOCATION_CHANGE = '@@router/LOCATION_CHANGE'
```
gaearon: I think we need more descriptive names so the difference between UPDATE_LOCATION and LOCATION_CHANGE is obvious. It particular we want to make it clear that only intercepting LOCATION_CHANGE is guaranteed to give you every navigation event. We might want to rename UPDATE_LOCATION to REQUEST_LOCATION_CHANGE, for example.

webmasterkai: or even `INVOKE_LOCATION_METHOD`

timdorr: I'd like to leave the UPDATE_LOCATION action in place to reduce some of the upgrade burden on those who will continue to use the middleware. This one can be renamed to whatever we like.

gaearon: I'm not sure keeping the old name is worth the confusion generated by introducing two actions.

Especially now that we encourage calling history methods directly, people relying on the old action and thinking they don't need to change their analytics code will have broken code because direct history calls will not invoke the new action.

I think CALL_HISTORY_METHOD might be the best choice here. It makes it clear this is not intended for analytics.

webmasterkai: What about an export alias with note in documentation about deprecation/change?

webmasterkai: Made a pull request #265 for this change.

webmasterkai:
One last point on setting initialState and then I'm (probably) done forever.

Hitting the "Reset" button on DevTools reverts to the initial state of the store. It wipes out the first action and the result of its associated payload. I don't know about @gaearon, but when I click "Reset" I expect the location state to return to what it was at the time the store was created. I find it unacceptable that if I enter the app with `/foo`, navigate to `/bar` and then click "Reset" my address bar changes to `/`. :dizzy_face: 

My proposed solution is exporting a utility function like the following:

```javascript
export function getInitState(history) {
  let locationBeforeTransitions
  history.listen(loc => locationBeforeTransitions = loc)()
  return { locationBeforeTransitions }
}
```

With usage like this:
```javascript
import { createStore, combineReducers } from 'redux'
import { routerReducer, routerMiddleware, getInitState } from 'react-router-redux'
const history = createHistory()
createStore(
  combineReducers({
    routing: routerReducer
  }),
  {
   routing: getInitState(history)
  }
)
syncHistoryWithStore(history, store)
```

Any objections or improvements? :ear:

timdorr:
@webmasterkai That's not what happens. Depending on how you flip an option, the library can sync the url when state changes. It will do that by default and when you reset, the url updates appropriately.

src/sync.js
gaearon: Now that I looked it up, Rx 5 will actually use `unsubscribe()` too so `dispose()` doesn’t make sense as more obscure naming.

Since `syncHistoryWithStore` does not *create* the underlying history, it shouldn’t try to destroy it. Therefore my original comment doesn’t make sense, and we should leave this as is. :+1:

webmasterkai:
@timdorr I'm not talking about toggling actions on/off. I am talking about clicking the "Reset" button. Upper left of DevTools pane.

timdorr:
That's what I'm referring to too. It works in my testing with the basic example.

webmasterkai:
Steps to reproduce out of sync window.location/redux store:

1. Open the /example/basic code in a browser.
1. Enter the app with the address http://localhost:3000/#/foo
1. Click the "Bar" link
1. Click the DevTools "Reset" button.
1. Store shows `null`. Address bar shows `/#/foo`.

src/sync.js
```diff
@@ -0,0 +1,136 @@
...
+  let initialLocation
+  let currentLocation
+  let isTimeTraveling
+  let unsubscribeFromStore
+  let unsubscribeFromHistory
+
+  // What does the store say about current location?
+  const getLocationInStore = (useInitialIfEmpty) => {
+    const locationState = selectLocationState(store.getState())
+    return locationState.locationBeforeTransitions ||
```
webmasterkai: Unless we expose `getLocationInStore()` this is private state management and results in window.location and the redux store not mirroring one another.

webmasterkai:
How about sending history to a `createReducer` function?

```javascript
export function createReducer(history) {
  const initialState = {}
  history.listen(loc => initialState.locationBeforeTransitions = loc)()

  return (state = initialState, { type, payload }) => {
    if (type === LOCATION_CHANGE) {
      return { ...state, locationBeforeTransitions: payload }
    }
    return state
  }
}
```

timdorr:
It wouldn't be a pure function. Reducers must be pure functions.

webmasterkai:
It's a higher order function that returns a reducer. Just like `combineReducers()` returns a reducer.

timdorr:
Ah, I didn't read too closely. I don't think we're going to do that though. I'm mobile, so I can't check what's going on in your case. But the fact you're loading up an server says something is wrong, because the basic example has no server. You load it up from the index.html file.

webmasterkai:
What about #6 and #207 and the fix https://github.com/rackt/react-router-redux/commit/a00acfd4310ba5eaf3a9530b0f3fa8bd3d312c6e?
Why have these issues suddenly become unreasonable feature requests?

timdorr:
- #6 is old and doesn't work without a browser.
- #193 is unrelated and should be sent through `history.listen`.
- #207 / a00acfd rely on the 2.0/3.0 structure, so they don't apply as a fix here.

I wasn't able to repro your issue with the URL against the basic example. It resets to the correct location via the internal `initialLocation`. 

This is fine, as we are essentially doing the same as devtools does when it lifts the state to store a copy of the `initialState` for reset. We are not modifying anything internal with the store or doing anything funky with updating `initialState` just-in-time. We are simply replaying the initial POP action that occurs when you reload the page. This is inline with devtools' reset, which represents reloading your application from scratch. It fits semantically, so the behavior is correct.

gaearon:
I have not tested this yet. But:

>Store shows null. Address bar shows /#/foo.

This does not seem right. What I would propose is to check location in store inside the store change handler. If location is null, we need to dispatch the initial action with the current history value. This way initialisation would automatically occur on every reset. 

Does this make sense?

timdorr:
Here's what it looks like when loaded:

screen shot 2016-02-06 at 6 14 16 pm

And here's what you get when you immediately hit the reset button:

screen shot 2016-02-06 at 6 14 29 pm

Same when you navigate around and then hit reset. I swore I was getting that initial replay done. I'll do some fiddling, but the tricky part will be not getting into an infinite loop.

webmasterkai:
@gaearon yes that would work, but dispatching an action is treating the symptom instead of fixing the cause. The symptom is that `null` is not a valid representation of the current window.location.

"Oops, locationBeforeTransitions is `null`. Better dispatch an action to fix it."

@timdorr as your graphic clearly illustrates `null !== /foo`.

gaearon:
Can you please describe the practical problem with "dispatching an action is treating the symptom"?
I understand the ideological stance on this but I want to see if there are any practical issues here as well.

In my point of view the late binding of history is a design decision. We don't want createReducer(history) because we don't want it to be a singleton. Asking user to provide initial state to the store just for this also seems sketchy. Therefore embracing "null means this history is not yet associated with this store, and needs to be if we're inside syncHistoryWithStore" seems like the most meaningful option, even though I understand it's not as elegant as we would like.

timdorr:
> It's another dispatching an action after everything has mounted.

This event will dispatch before the mounting has occurred, as it happens synchronously inside of `syncHistoryWithStore` and you have to provide the return value (the enhanced `history`) to the `<Router>` before render.

webmasterkai:
@gaearon If a component depends upon the store having a valid representation of location and then it's null the component is going to puke. Yes, the component could handle not having a null location value. It also enables sync.js to slim by at least 18 lines instead of having to **add lines** to check for null then dispatch. I agree that my stance is largely ideological.

Dispatching an action before any component is mounted I can live with. Dispatching an action as a result of the store being incorrect after components have mounted I can't.

@timdorr I have no reason to use <Router>.

Thank you all for the discussions. It's been an exciting part of my day.

gaearon:
>It also enables sync.js to slim by at least 18 lines instead of having to add lines to check for null then dispatch

If we were taking decisions based on how much lines they take this library wouldn’t exist at all :wink: . It’s all about the user and developer experience.

>If a component depends upon the store having a valid representation of location and then it's null the component is going to puke. 

OK, this starts to make sense to me. Are you saying that you don’t use `react-router` itself and plan to read from `state.routing.locationBeforeTransitions` yourself?

In this case, I indeed see the issue. With React Router you wouldn’t really care about that `null` because it would not be used in the UI anyway, as [we are passing `useInitialIfEmpty = true`](https://github.com/rackt/react-router-redux/blob/a4527911eacb283073a5b35a0106ac649191a25b/src/sync.js#L111). However if you intend the UI to read from `state.routing` directly, it would throw unless you add a precaution.

I am not 100% you would hit this scenario in the real life though. If you call `syncHistoryWithStore` right after the store was created, the listener that dispatches the synchronous action would be in the beginning of the listener list. Redux invokes listeners in the order of their subscription. This means that whatever UI is connected to `state.routing`, by the time its listeners are called, `state.routing` should be filled by that time. I guess this is not as neat as you would like it to be, but it seems that the problem is averted, and on a broader scope, this library’s primary goal is supporting React Router workflows rather than vanilla history so even though we technically support your pattern of usage, we don’t want to inconvenience RR users because of an issue they will never hit.

webmasterkai:
> ...this library’s primary goal is supporting React Router workflows rather than vanilla history.

@gaearon I think you're on point here. I have no plans to use React Router and ultimately decided to go in a completely different direction.

Thanks again and cheers to all!

gaearon:
>I have no plans to use React Router and ultimately decided to go in a completely different direction.

You can still use this lib fine though.
All I’m saying is we don’t want to add inconveniences for the RR users for a theoretical gain.
Especially considering your scenario should work fine as well:

>Redux invokes listeners in the order of their subscription. This means that whatever UI is connected to state.routing, by the time its listeners are called, state.routing should be filled by that time.

If I am wrong please correct me, but I think the approach with dispatching an initial action should solve your issue too.

timdorr:
OK, I'm all done with this. I'll probably publish this as a beta release under the `next` dist-tag sometime in the next day or two, provided no one sees any huge issues.

src/sync.js
```diff
@@ -0,0 +1,136 @@
...
+ * returned with a listen method that responds to store updates for location.
+ *
+ * When this history is provided to the router, this means the location data
+ * will flow like this:
+ * history.push -> store.dispatch -> enhancedHistory.listen -> router
+ * This ensures that when the store state changes due to a replay or other
+ * event, the router will be updated appropriately and can transition to the
+ * correct router state.
+ */
+export default function syncHistoryWithStore(history, store, {
```
SimenB: The third argument is not mentioned in the docs at all, AFAICT

timdorr: Added in 0b7b322. Thanks for noticing!

timdorr:
Published on the `nest` dist-tag. Install it and give it a whirl!
```sh
npm install react-router-redux@next
```

ezekielchentnik:
guys, i'm way behind on all this... but want to contribute.

the way i see it, react-router-redux does not need to be glued to react-router, although, i realize, 90% of the time, it will be. 

really, the point is, history/routing is just more state that should exist in the store.... can we officially declare this concept? (in docs, or examples)?

the way I see it, redux is a pattern that can be evolved independently of react; that said, the reason I forked and created 'redux-history' was for these sentiments.  'redux-history' in it's name holds a lot of value and direction.  I think we can take advantage of that and increase adoptability and evolve...

the current reality is that syncing with redux store and vice versa can happen completely independent of react-router; however, that is not very clear to the average consumer... I'm talking about those only reading docs and not examining deep code.

maybe I'm off my rocker, let me know.  I'd love to hear opinions/thoughts - but one of the reasons i was attracted to the OG 'redux-simple-router' was after I tried several other modules, read the code, and realized I don't have to care about react-router and it fit my use of stuffing location into the store.  I like that, and it's simplicity - its about redux, not react.

gaearon:
I agree I would have preferred this to be called redux-history.
I would release project as is and reevaluate whether we should keep the name in a couple of months.

timdorr:
@ezekielchentnik I had this to say before: https://github.com/rackt/redux/pull/1362#issuecomment-179929409

> On the naming part: React Router 2.0 hides the History project a little better. You're still talking to history objects, but they're provided via the Router's APIs. And there's no peerDep on history. So, it's visibility is going away and I think most people, from a "brand" perspective, are going to associate this with React Router vs History.

`history` is going more commonly observed as a React Router feature in 2.0, even if it is a separate library. However, most people won't know that and, frankly, won't care. They'll just `npm install react-router` and it comes along for the ride. 

This library isn't tied to React Router by code (it's not in the deps), but there is a connection by use case. The vast majority are using this to connect to React Router. The examples we give assume use of React Router. The issues brought up almost always involve React Router. It's always there. As such, I think the name fits well.

SimenB:
Any plans to make this IE8 compatible? The exports use getters now

romseguy:
Sorry for jumping in without reading the whole thread but I gave 4.0.0-beta.1 a spin along with RR2.0.0-rc6. I'm using SSR and accessing e.g `/login` is correctly matched and rendered on the server, but I am redirected to `/` shortly after. Accessing the route from `view-source:http://localhost:3000/login` here's the store state I got from the server:

```
{
  "routing": {
    "locationBeforeTransitions": {
      "pathname": "\/",
      "search": "",
      "hash": "",
      "state": null,
      "action": "POP",
      "key": "0t38e1",
      "query": {
        
      },
      "$searchBase": {
        "search": "",
        "searchBase": ""
      }
    }
  }
}
```

I'm not familiar with the internals but it seems the store isn't synced with history state after `/login` is matched.

My (trimmed down) implementation:

```javascript
// server.js
import { createMemoryHistory } from 'react-router'

const { store, syncedHistory } = createStoreWithHistory(createMemoryHistory())
const location = history.createLocation(req.url)
match({routes, location}, (error, redirectLocation, routingProps) => {/* ... */})
```

```javascript
// client.js
import { browserHistory } from 'react-router'

const { store, syncedHistory } = createStoreWithHistory(browserHistory, window.__data)
```

```javascript
// createStoreWithHistory.js
import { syncHistoryWithStore } from 'react-router-redux'
import { createStore } from 'redux'

function createStoreWithHistory(history, initialState) {
  const store = createStore(reducer, initialState)
  return {store, syncedHistory: syncHistoryWithStore(history, store)}
}
```

timdorr:
@SimenB I can look at IE8 compatibility and implement Redux's solution.

Yes, you can keep using the plain history singleton from React Router. The only difference between the two is the `listen` implementation, so `push` and the rest of it is the exact same function.

@romseguy I haven't done any SSR testing yet. But try this instead:
```js
import { createMemoryHistory } from 'react-router'

const { store, syncedHistory } = createStoreWithHistory(createMemoryHistory())
match({routes, location: req.url, history: syncedHistory}, (error, redirectLocation, routingProps) => {/* ... */})
```
You definitely need to pass the enhanced history into `match`, otherwise it doesn't get used in the router.

romseguy:
Same problem.

svrcekmichal:
Having the same issue as @romseguy , created this issue #284

timdorr:
@romseguy & @svrcekmichal This was fixed by properly seeding our memory history with the right path:
```js
const memoryHistory = createMemoryHistory(req.path)
```
I'll have to check that our SSR docs on react-router are consistent.

timdorr:
I haven't seen any serious issues come up, so I'm going to merge this into master. I've cut a `3.0.x` branch and will leave that as default for a while until 4.0.0 goes final. 

While there haven't been any changes to the code, I'm going to push an rc1 release shortly. Hopefully, that will signal to more people to give it a try. It works well! That should root out some further bugs and we can make this thing about as good as it can be.

romseguy:
Thank you @timdorr and everyone else, works great now.

transtone:
how to use it, is there a doc?

SherylHohman:
Yes, please Do rename it to redux-history !

PR #262, opened 2016-02-06
Following feedback in #259

webmasterkai:
See issue #259.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ View the [CHANGELOG](https://github.com/rackt/react-router-redux/blob/master/CHA
 
 Read the [API docs](#api) farther down this page.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -73,7 +73,7 @@ ReactDOM.render(
 )
 ```
 
-Now you can read from `state.routing.location.pathname` to get the URL. It's far more likely that you want to change the URL more often, however. You can use the `push` action creator that we provide:
+Now you can read from `state.routing.locationBeforeTransitions.pathname` to get the URL. It's far more likely that you want to change the URL more often, however. You can use the `push` action creator that we provide:
```
timdorr: User are discourage from reading state at all. I'm thinking of prefixing this key with a `$` to make that clear. But this section of the README should actually go away or be significantly changed.

timdorr:
Cool, I'll merge this back into the branch so it shows up in the PR for further work.

README.md
webmasterkai: What's the solution for users that refuse to define routes in jsx within a react component? Where should a user parse the location so they can get things like named path parameters using a tool like url-pattern?

timdorr: Router hasn't had named routes for a while now. You should be using URLs or location descriptors for all of your navigation.

webmasterkai: I'd like to make a simple example that does not use react-router. How do I gain access to the location.pathname?

timdorr: Listen to the history. It's a synchronous operation.

webmasterkai: Isn't that the point of this library? To listen to history and put it in the same place where every other component reads state from?

PR #263, opened 2016-02-06
Re: #259 Small tweaks to documentation

webmasterkai:
Merge if you feel like it. Incremental changes getting closer to describing the future. :-)
Following #259 
Basically just trying to get comments into the README so it's there during a rewrite.


timdorr:
I just redid the whole readme, so this won't merge in cleanly. Any further changes should be based on that version.

PR #265, opened 2016-02-06
rename UPDATE_LOCATION to CALL_HISTORY_METHOD

webmasterkai:
Following comments on #259.

> I'm not sure keeping the old name is worth the confusion generated by introducing two actions.
>
>Especially now that we encourage calling history methods directly, people relying on the old action and thinking they don't need to change their analytics code will have broken code because direct history calls will not invoke the new action.
>
>I think CALL_HISTORY_METHOD might be the best choice here. It makes it clear this is not intended for analytics.
- @gaearon 

Also, let's not forget the the name already changed once in the past... 
> * `UPDATE_PATH` is now `UPDATE_LOCATION`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,14 +83,14 @@ function MyComponent({ dispatch }) {
 }
 ```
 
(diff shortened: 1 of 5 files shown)


webmasterkai:
:kiss: Thank you!

Issue #268, opened 2016-02-10
listenForReplays is making async action in history.listenBefore never ending loop


svrcekmichal:
Hi, after little bit of hacking with react, redux, router and redux I came to strange issue, which i believe was bug in [history package](https://github.com/rackt/history/issues/227)

Then i tried to locate the problem and what i found out is that there is no problem with history, neither react-router implementation, but in react-router-redux package.

I tried to put down some functionalities and found out that issue is coming from `listenForReplays`. How to simulate problem an be seen in original issue  in history package.

I'm going to try and dig little more to found exact problem, but if there is someone who knows right now where' s problem it would be awesome

gaearon:
Mind giving `4.0.0-beta.1` a spin? I wonder if https://github.com/rackt/react-router-redux/pull/259 solved it. Note that it has a slightly different API: https://github.com/rackt/react-router-redux/blob/synchronicity/README.md

gaearon:
If it doesn’t solve the problem please provide exact ways to reproduce with the bundled example project.

svrcekmichal:
@gaearon going to try and reproduce, for now i was able to to stop the issue, but it has not so much to do with redux-react-router, issue stopped when i disabled dev tools, especially when i commented this [two lines](https://github.com/gaearon/redux-devtools-log-monitor/blob/master/src/LogMonitor.js#L118-L119)

if i dont comment those two lines, it works if i use setTimeout with value around 100ms, i guess even values with less than 1000ms will be working in special occasions, as it is lower than timeout in log monitor. If bigger value is used, you cant use router at all as all actions are overriden.

To see the issue for now, you can clone this [branch](https://github.com/svrcekmichal/universal-react/tree/feature/listen-for-replies)

For now, I'm going to try that `4.0.0-beta.1` and let you know how it works

gaearon:
Please do try the new branch. If there is an issue it *is* related to this library. The fact that log monitor dispatches actions is not relevant: that the library fails when actions are dispatched fast is the problem, and we need to figure out why, and check whether 4.x will fix that.

svrcekmichal:
@gaearon everything works, for now you are my little god :+1: I'm going to check what you have change to learn, I have so much to learn from you :) thanks once more

svrcekmichal:
[...]

PR #269, opened 2016-02-11
only update on location changes

Autarc:
Currently a `replace` action is forced on any store dispatch at the initial route. Aside of polluting the log its also a problem in case `history.listenBefore` is used asynchronously. It triggers the listeners although the route didn't change.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -80,18 +80,20 @@ export function syncHistory(history) {
       unsubscribeStore = store.subscribe(() => {
         const location = getLocationState()
 
-        // If we're resetting to the beginning, use the saved initial value. We
-        // need to dispatch a new action at this point to populate the store
-        // appropriately.
-        if (location.key === initialLocation.key) {
-          history.replace(initialLocation)
-          return
-        }
-
-        // Otherwise, if we need to update the history location, do so without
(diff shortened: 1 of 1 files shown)


timdorr:
Can you write up a small test to accompany this? Or do you have a use case I can look at?

Autarc:
Checked if writing a test would be reasonable, but ended with renaming the existing [handles toggle after store change](https://github.com/reactjs/react-router-redux/blob/master/test/createTests.js#L256) into `handles toggle after routing store change` which seems redundant. The issue is that each time the store gets updated the `replace` action can occur even at the initial page, not only when there is an actual change at towards `routing` but whenever an action gets dispatched.

This leads to errors regarding `beforeListen((location, callback => ...)` which will be triggered multiple times if an action is dispatched to set data for the next page. You can find [here](https://github.com/Autarc/redial#example-client-usage) an example with asynchronous fetching at the client.

react-router-redux-keepinitiallocation

svrcekmichal:
Have you looked at #268 ? I was having the same issue, it was fixed with in new version.

timdorr:
This is going to be superseded by 4.0.0, so I would upgrade to the rc.1 release and see if that works better for you. If not, open an issue on that version and we'll see about getting it fixed.

PR #280, opened 2016-02-15
added example to readme

svrcekmichal:
Added another one example to new implementation


.gitignore
```diff
@@ -2,3 +2,4 @@ lib
 node_modules
 coverage
 *.log
+.idea
```
timdorr: Can you get rid of this? You should add this to your global .gitignore

svrcekmichal: @timdorr amended, didn't noticed, thanks :)

timdorr:
Thanks!

PR #281, opened 2016-02-15
Add example

quicksnap:
We've finally migrated in master, thanks to @sars!


timdorr:
Thanks and congrats! :smile:

PR #289, opened 2016-02-18
Fix issue when initial location redirects

decafdennis:
We've been using the cutting edge version in a new project, and all work that went into v4 is much appreciated!

Today we found a bug: if the initial route redirects to another location, then the `Router` component doesn't update properly and leaves the user with an empty page. It's common to use `onEnter` on a `Route` to redirect the user to a login page when she is not logged. The `syncs history -> components when the initial route gets replaced` test in this PR replicates the problem.

The problem lies in the `syncHistoryWithStore`'s custom listening logic. It makes the first synchronous call to each listener before it subscribes to the store changes, so if the listener causes a location change and thus a store change (e.g. when a location redirects), then the listener is not notified of that change. This PR fixes the problem by subscribing to the store before making the first call to the listener.

Hopefully this bug can be fixed before the final v4 is released!

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "mocha": "^2.3.4",
     "react": "^0.14.3",
     "react-dom": "^0.14.3",
+    "react-redux": "^4.4.0",
     "react-router": "^2.0.0",
     "redux": "^3.0.4",
     "redux-devtools": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -64,6 +64,8 @@
     "karma-webpack": "^1.7.0",
     "mocha": "^2.3.4",
     "react": "^0.14.3",
+    "react-dom": "^0.14.3",
+    "react-router": "^2.0.0",
     "redux": "^3.0.4",
     "redux-devtools": "^3.0.0",
     "redux-devtools-dock-monitor": "^1.0.1",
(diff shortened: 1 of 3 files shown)


schickling:
Hit the same problem. Thanks for creating this PR! :shipit:

btmills:
Thanks for this @decafdennis!

For those who need a workaround until this fix is released, you can use the asynchronous callback:

```js
function requireAuth(nextState, replace, callback) {
    if (!isAuthenticated()) {
        replace({
            pathname: '/login',
            state: {
                nextPathname: nextState.location.pathname
            }
        });
    }

    setTimeout(callback, 0); // TODO: Remove this once https://github.com/reactjs/react-router-redux/pull/289 is released
}
```

decafdennis:
Thanks for the workaround @btmills!

I also published `@decafdennis/react-router-redux` to NPM as a temporary workaround. It's 4.0.0-rc.1 with the fix. You can alias the package name in your `webpack` config if you're using that.

timdorr:
Interesting! Thanks for writing a solid test case for this too.

schickling:
Awesome! Will there be a new version for this?

timdorr:
Yeah, I'll do a quick rc2 tonight.

schickling:
Thanks @timdorr! :+1:

decafdennis:
@timdorr: thanks for the quick merge!

PR #304, opened 2016-02-24
Fix the basic example to use the same history everywhere

nhurman:
https://github.com/reactjs/react-router-redux/blob/master/examples/basic/app.js#L30

The store is synchronized with browserHistory, but the App component uses hashHistory.


nhurman:
Fix already available at https://github.com/reactjs/react-router-redux/pull/303

PR #309, opened 2016-02-25
Update README.md

StevenIseki:
add link to react-router-redux-example - minimal react-router-redux example includes css modules and universal rendering


timdorr:
Thanks!

PR #310, opened 2016-02-26
Fix `LOCATION_CHANGE` action on README

fdaciuk:
The correct action is `@@router/LOCATION_CHANGE`: https://github.com/reactjs/react-router-redux/blob/master/src/reducer.js#L5

=)


timdorr:
Nope, that's value. [It's exported as `LOCATION_CHANGE`](https://github.com/reactjs/react-router-redux/blob/master/src/index.js#L2)

fdaciuk:
Oh, sorry :x

I was trying to use the string format, instead to import it.

Thanks =)

PR #329, opened 2016-03-13
Add example to README.md

kuy:
An example for react-router-redux with d3's treemap.

preview


timdorr:
Oh, that's super-neat! Thanks!

PR #334, opened 2016-03-20
Document: improve "LOCATION_CHANGE action type"

SimplyY:
Oh I see it has something can be improved in doc. Because it(LOCATION_CHANGE) changed names since 3.0.0（in this year）, so doc should  show it.


timdorr:
The library completely changed in 4.0.0, so there isn't an upgrade path and we aren't documenting changes. Thanks for this anyways!

PR #342, opened 2016-03-29
Explanation for needing to pass children property

harunsmrkovic:


timdorr:
This isn't needed. This is a react-router concern.

PR #344, opened 2016-03-31
Fix compatibility issue with IE8 (ES3) due to usage of unquoted reserved keys.

### LINKED ISSUE #332: Not compatible with IE8 (ES3)
The output in `./lib` is not compatible with IE8 due to usage of unquoted reserved keys (e.g. `export.default` should be `export['default']`). I forked the repo hoping that I could fix it using `babel-plugin-transform-es3-property-literals` and `babel-plugin-transform-es3-member-expression-literals` but as this thread indicates it doesn't solve the issue: https://phabricator.babeljs.io/T2817. It seems using `es3ify` is a workaround for now. I can transform your package at build time on my end, but I wonder if it's something that should be addressed here instead.

ctumolosus:
Issue was fixed in Babel v6.7.4
https://github.com/babel/babel/releases/tag/v6.7.4

Closes #332


timdorr:
Thanks! I'll push this out shortly.

PR #350, opened 2016-04-05
Fix route reducer crash

### LINKED ISSUE #349: Calling routerReducer() with no args crashes
If you call `routerReducer()` with no arguments to get back the initial state, it crashes with the following error:

`reducer.js: Uncaught TypeError: Cannot read property 'type' of undefined`

The generated code looks like this:

    function routerReducer() {
      var state = arguments.length <= 0 || arguments[0] === undefined ? initialState : arguments[0];
      var _ref = arguments[1];
      var type = _ref.type;       // <==== _ref is undefined here and _ref.type causes the error
      var payload = _ref.payload;
    
      if (type === LOCATION_CHANGE) {
        return _extends({}, state, { locationBeforeTransitions: payload });
      }

      return state;
    }

This can be fixed in the original code by adding a default parameter for the action parameter:

    export function routerReducer(state = initialState, { type, payload } = {}) {

It made me wonder though what the correct behavior of Babel should be here.  Should it provide safe destructuring or still leave that up to your code?  I checked Traceur though and it behaves the same way, so perhaps that's not part of the spec, but it seems dangerous.

mindjuice:
Fixes #349 in parent

Also fixed a typo in the function comment.


timdorr:
Thanks!

PR #351, opened 2016-04-07
Update the README to reflect changes to react-router

hdh3000:
updated the readme to reflect that react-router has split off a seperate history package, and that you should actually call 'history.createHistory' instead of importing the now non-existent browserHistory


timdorr:
It's correct for Router 2.0. Make sure you're looking at the right version. You've got things built for 1.0.

PR #363, opened 2016-04-17
Update ImmutableJS instructions

ericrav:
Took me a while before I realized I needed to convert my Immutable object to a plain object for this to work, so this may help others


timdorr:
Thanks!

PR #387, opened 2016-05-02
Clarify the goal of this library

gaearon:
This isn’t very clear right now, and causes some confusion.


timdorr:
Seems OK to me 👍

PR #403, opened 2016-06-02
Init currentLocation to initial location from the store

### LINKED ISSUE #313: Need to click browser back button twice to exit page in SSR example
I'm following the SSR example here:
https://github.com/reactjs/react-router-redux/blob/master/examples/server

Open [http://localhost:8080](http://localhost:8080), click browser back button, redux will dispatch router/LOCATION_CHANGE action, the page will remaining in same [http://localhost:8080](http://localhost:8080). Click browser back button again, then only the page will be close. 
I noticed same issue on other community example as well such as https://github.com/svrcekmichal/universal-react

dlmr:
This solves the problem when the server has initialised the state and the user already has things in their history.

Without this change a superfluous action will be pushed with the same URL that the user is already looking at resulting in that we would need to use the back button twice.

Resolves #313

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -31,7 +31,6 @@ export default function syncHistoryWithStore(history, store, {
   }
 
   let initialLocation
-  let currentLocation
   let isTimeTraveling
   let unsubscribeFromStore
   let unsubscribeFromHistory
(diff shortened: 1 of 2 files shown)


timdorr:
Thank you for this! Sorry for the delay on the review. I'm prepping our house for our first kid, so things have been crazy at home. And I wanted to give this the right attention. I'll push out a new version shortly.

PR #416, opened 2016-06-28
Improve instructions for using with Immutable.js

### LINKED ISSUE #415: Clarify usage with Immutable.js
In the readme, it says

> Provide your own reducer function that will receive actions of type LOCATION_CHANGE and return the payload merged into state.

This isn't enough information, and the user needs to look through the source code to understand where to actually put the data.

Can we change this to something like

1. Provide your own reducer function that will receive actions of type LOCATION_CHANGE and return the payload merged into the `locationBeforeTransitions` property of the routing state. e.g. `state.set("routing", {locationBeforeTransitions: action.payload})`.

sfentress:
1. It was easier for me to think of the reducer by first knowing where
the state was expected to be. Writing it the other way around was
confusing
2. Add that the action payload data must go into the 
`locationBeforeTransitions` property

Closes #415

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,12 +81,12 @@ history.listen(location => analyticsService.track(location.pathname))
 
 For other kinds of events in your system, you can use middleware on your Redux store like normal to watch any action that is dispatched to the store.
 
-#### What if I use Immutable.js with my Redux store?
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

PR #427, opened 2016-07-19
Fix ie8 compatibility issue

### LINKED ISSUE #382: Fix IE8 Compatibility
Ref #379

Right now, as bundled via Babel 6, our named exports (`export foo from './foo'`) generate code that breaks in OldIE (at least IE8). There is a pending bug open with Babel to fix this: https://phabricator.babeljs.io/T7322

andyhu:
Fix #382

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,9 +1,15 @@
-export syncHistoryWithStore from './sync'
-export { LOCATION_CHANGE, routerReducer } from './reducer'
-
-export {
+import syncHistoryWithStore from './sync'
+import { LOCATION_CHANGE, routerReducer } from './reducer'
+import {
   CALL_HISTORY_METHOD,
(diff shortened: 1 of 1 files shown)


timdorr:
You shouldn't be using this in production. This is primarily for dev tools support only. I don't think anyone is actively developing on ie8.

andyhu:
But it doesn't hurt to merge it right? There's nothing to lose by supporting IE8?

mjrussell:
> You shouldn't be using this in production.

@timdorr Im not sure I would agree with that statement. Even if you aren't using `syncHistoryToStore`, it seems perfectly legitimate to use `routerMiddleware` to dispatch history actions in production.

timdorr:
We only have those as a convenience. I'm planning on removing them in the next release to encourage users to use the history singletons in React Router instead.

Issue #444, opened 2016-08-26
After initial render first action requires two clicks


pitr12:
Hello guys!

In our App where we are using react-router-redux, we experienced the same problem as described in [313](https://github.com/reactjs/react-router-redux/issues/313). New release regarding to [403](https://github.com/reactjs/react-router-redux/pull/403) fixed this problem as it was supposed to, but it caused new problem in our app that forced us to downgrade back to previous release.

New problem that [403](https://github.com/reactjs/react-router-redux/pull/403) is causing is that after initial render first action requires two clicks. For example if I refresh page on landing page and then click some menu item which is supposed to redirect me to new location, nothing happens and than if i click again after that I am redirected to new location. (To be more precise, after first click I am redirected to location at which i actually am. Eg. I am on landing page, hit refresh and after refresh i click on menu item 'about', and my location is replaced to same location 'landing' and after second click on item 'about' location is changed correctly).

Issue #429, opened 2016-07-29
Double location change pushing an url in action creator


maiopirata:
I've an Isomorphic app.
I'm experiencing a problem similar to #403 I tried to put in my package.json:
"react-router-redux": "dlmr/react-router-redux#fix-inital-server-load-dist", 
but it is not working...

In my action creator I have:
```
export function formComplete(id, router) {
  return (dispatch, getState) => {
    dispatch(checkForm(id))
    
    if (getState().registration[id]) {
      return dispatch(formPassed(id, router))
    }
  }
}
```
The first action check that the page is completed and in that case a second action is dispatched:
```
export function formPassed(id, router) {
  router.push('/registration')
}
```

Now I'm in an url that is `/registration/login`. When I click to validate the form the two actions are dispatched in the correct order and finally my app redirects to `/registration`.. But it automatically redirect a second time to `registration/login`

My client.js:
```
let preloadedState = window.__PRELOADED_STATE__
const persistedState = loadState()
if (persistedState != null) preloadedState = persistedState

const store = configureStore(preloadedState, browserHistory)
const rootElement = document.getElementById('app')
const history = syncHistoryWithStore(browserHistory, store)

store.subscribe(throttle(()=> {
  saveState(store.getState());
}, 1000))

render(
  <Provider store={store}>
    <Router routes={ routes } history={ history } 
      onUpdate={() => { window.scrollTo(0, 0) }} 
     />
  </Provider>, rootElement
)

```
my server.js
```
function storeUpAndGo (sessionState, history) {

    // setup of the Redux store
    let preloadedState = { rootReducer }
    const persistedState = loadState(sessionState)
    if (persistedState != null) {
      preloadedState = persistedState
    }

    // Create a new Redux store instance
    return configureStore(preloadedState, history)

  }

  // This is fired every time the server side receives a request
  const jsFile = parameters.chunks().javascript.main
  const cssFile = parameters.chunks().styles.main

  app.use(handleRender)

  function handleRender(req, res) {

    const memoryHistory = createMemoryHistory(req.url)
    const store = storeUpAndGo(req.session, memoryHistory)
    const history = syncHistoryWithStore(memoryHistory, store)

    match({ history, routes, location: req.url }, (err, redirectLocation, renderProps) => {
      if (err) {
        console.error(err);
        return res.status(500).end('Internal server error');
      }

      if (!renderProps) {
        return res.status(404).end('Not found.');
      }

      const params = qs.parse(req.query)

[...]

PR #445, opened 2016-08-26
Makes sure the state in the store matches the state in history when using SSR

dlmr:
This solves some problems where they would be out of sync on the first render when using server side rendering. The location would be the same for both of them but the `key` would be different resulting in some strange behaviour.

The application will now always dispatch a `LOCATION_CHANGE` on initial render. This was already done when not using server side rendering, meaning that the behaviour is similar between using server side rendering and not.

I have tested this manually and it seems to work fine. I have unfortunately not  enough time right now to add more tests.

I believe that this PR supersedes #412 and will probably address #444 and #429.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -34,6 +34,7 @@ export default function syncHistoryWithStore(history, store, {
   let isTimeTraveling
   let unsubscribeFromStore
   let unsubscribeFromHistory
+  let currentLocation
 
   // What does the store say about current location?
   const getLocationInStore = (useInitialIfEmpty) => {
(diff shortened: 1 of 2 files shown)


ptomasroos:
Any news?

dlmr:
I don't have time at the moment to write tests but I believe the implementation is solid. Help here is appreciated.

unRob:
@dlmr These changes work for our project too, enabling the transition from server to client without a hitch. I'd love to have this PR land soon, could you help me figure out what the approach to tests should be so I can contribute with some?

timdorr:
OK, let's do it!

timdorr:
Out in 4.0.6. Sorry it took so long!

PR #446, opened 2016-08-27
README.md: move applyMiddleware to it's respective example

viatsko:
It's not used in 1st example, but used in the 2nd one.

This change makes it clear that following imports are coming from redux in the 2nd example, while cleaning up unused import in 1st.


timdorr:
Thanks!

PR #476, opened 2016-10-25
Support history 3

### LINKED ISSUE #475: Compability with History v3
React router v3 was released today. The new version also updates dependency for history package. Now it requires history@3.x.

Then it turns out, that history has [a breaking change](https://github.com/mjackson/history/blob/v3.0.0/CHANGES.md#300-0)
> Breakage: history.listen no longer calls the callback synchronously once. Use history.getCurrentLocation instead

We need to detect new version and handle it properly.

just-boris:
Fixes #475.

The idea how to determine new version of history I took from `react-router` [source code](https://github.com/ReactTraining/react-router/blob/master/modules/Router.js#L79)


timdorr:
That was easy 😄 

Lemme test a few things and then I'll release this, provided it works OK.

dlmr:
Is there anything I can help with to release a new version that adds this feature?

yoadsn:
Updating to latest react-router (3.0.0) and latest react-router-redux  (4.0.6) breaks with async ssr use-case. The state is sent to the client with an empty location (expected?) and no location is found on state upon calling the "match" again - this throws since location is expected in react-router for "match" to work.

This is most likely due to the mismatch with history 3.0.0 - and not getting the current location from it. (no auto "listener" call on 3.0.0).

I think it's worth trying again with this PR applied - but still waiting for it to land on npm.
(Currently don't have time to build locally and test).

If this is pending for some reason and I can help with something (problem repro, testing local build) let me know and I'll try to clear up time for it.

timdorr:
Pushed out 4.0.7

yoadsn:
This solved the problem for me.
Thanks!

levity:
Upgrading major versions of history and react-router is a point release!? are you serious? going from 4.0.6 to 4.0.7 broke my server-side rendering setup...

PR #495, opened 2016-12-18
Readme: Update rackt references to point to /reactjs/...

spen:
Hi All 👋

Just a tiny one here!

I noticed a bad link in another project, linking to github.com/rackt/redux which is now held by somebody else it seems - who also has a redux repo, just to confuse things.

I opened a PR in that project but thought I'd have a quick look at other occurrences and noticed that there are a few vestiges here & updated them.

Ta 😀

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ These two hooks will allow you to store the state that this library uses in what
 
 #### How do I access router state in a container component?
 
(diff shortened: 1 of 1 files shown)


timdorr:
Ugh, thanks for the heads up. I'll double-check our other repos too. Thanks!

spen:
No problem :) Thanks for the review! The other repos should be good - always worth getting extra eyes on though!

Daniel15:
I'm surprised nobody picked up `rackt` as a placeholder when the org was merged with the `reactjs` one. Someone was bound to grab it at some point 😛

Issue #502, opened 2016-12-29
Wrapped history doesn't get updated


ForsakenHarmony:
if you wrap the history with `syncHistoryWithStore` it doesn't get updated anymore because you copy it [here](https://github.com/reactjs/react-router-redux/blob/master/src/sync.js#L107) 

This prevents a router that doesn't use `history.listen` from being able to get the correct value

timdorr:
What router doesn't use history.listen? That's kind of critical to how React Router works.

ForsakenHarmony:
preact-router

ForsakenHarmony:
Also i.e. history.location should still get updated

timdorr:
You need to pass in the history instance you get back from the syncHistoryWithStore function into your router.

ForsakenHarmony:
`history.location` doesn't even get updated in the wrapped history

ForsakenHarmony:
maybe something like this instead of the current return statement

https://hastebin.com/hevivahaxi.js

ForsakenHarmony:
https://hastebin.com/ucisiyidus.js something like this fixes it

ForsakenHarmony:
can you please reopen this at least?

ForsakenHarmony:
https://github.com/reactjs/react-router-redux/pull/504

PR #504, opened 2016-12-30
Fixes wrapped history not being in sync with the real one

ForsakenHarmony:
#502


timdorr:
We don't want to support history 4 because it is a rewrite and will go with a rewritten version of this library.

Also, this is a library for react-router. We aren't supporting alternative routers. This is only intended for time travel debugging, not production use.

ForsakenHarmony:
kinda funny, because this would have no downside for you

PR #510, opened 2017-01-16
Skip history call if the history location is currently the same as…

license2e:
… intended navigation location

Signed-off-by: Ilya Shindyapin <ilya@shindyapin.com>


timdorr:
You shouldn't pass through to the next action. 

You shouldn't be trying to capture duplicate actions in the middleware. That results in different behavior between dispatching an action and calling the method on history directly.

This doesn't work for the go*() methods.

PR #527, opened 2017-02-10
Fix test

rybon:
Fixes failing build on Travis CI.


timdorr:
Fixed this in another way. Double-pushing isn't actually correct here anyways.

PR #546, opened 2017-03-05
ESlint and scoping variable

hajjiTarik:


src/sync.js
```diff
@@ -37,8 +37,8 @@ export default function syncHistoryWithStore(history, store, {
   let currentLocation
 
   // What does the store say about current location?
-  const getLocationInStore = (useInitialIfEmpty) => {
-    const locationState = selectLocationState(store.getState())
+  const getLocationInStore = useInitialIfEmpty => {
+    let locationState = selectLocationState(store.getState())
```
alexrussell: @hajjiTarik how come you changed this `const` to a `let`? I don't see the value changing in the code.

timdorr:
Same here. Not really sure of the point. Thanks anyways!

PR #547, opened 2017-03-05
refacto updateLocation es6

hajjiTarik:


timdorr:
Thanks, but this doesn't really get us anything.

PR #554, opened 2017-03-22
Update README.md

krasevych:
add new starter kit with react-router-redux and other new technologies like Styled Components and Universal Webpack


timdorr:
Thanks!

PR #590, opened 2017-07-17
Sync: compare locations by structure

smoogly:
On first load `history.transitionTo` is called
even if state in store matches current history state
when `selectLocationState` doesn't preserve object reference
(for example using `Immutable.fromJS(...).toJS(...)`).

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -2,6 +2,14 @@ import { LOCATION_CHANGE } from './reducer'
 
 const defaultSelectLocationState = state => state.routing
 
+const isEqual = (state1, state2) => {
+  if (state1 === state2) {
+    return true
+  }
+
+  return JSON.stringify(state1) === JSON.stringify(state2)
+}
+
 /**
  * This function synchronizes your history state with the Redux store.
  * Location changes flow from history to the store. An enhanced history is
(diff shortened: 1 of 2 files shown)


timdorr:
Unfortunately, this isn't what we want to test on. We want to ensure we're dealing with the same location instance. Also, this would cause issues with `state` being set to something non-serializable.

smoogly:
My thinking behind this is: `selectLocationState` is user-supplied method. There's no guarantee it's going to preserve references.

Which is exactly the case in this project, for example: https://github.com/FullstackTypeScript/redux-bootstrap/blob/master/src/index.tsx#L69
(value is passed through Immutable.js; see this pen: https://codepen.io/anon/pen/dREXbX?editors=0010)

When `initialLocation` isn't equal to `locationInStore` by reference, extra history transition happens on first load and pollutes browser history (back button stops working properly).

So, may I ask you to clarify: is it the implementation incorrect or the problem should never be fixed in its entirety?

smoogly:
@timdorr, do you think being able to provide comparison function (which would default to reference comparison) is ok?
This way if `selectLocationState` for some reason doesn't preserve references, it will be possible to specify custom comparison to resolve this.

PR #591, opened 2017-07-19
Prevent undefined action crash

alvaro-paco:
I've done a little modification to prevent a undefined Exception. When you dont have signed action var in the middleware, the app crashes.


timdorr:
The action always needs to exist. If you don't pass one into dispatch, that's not using Redux correctly. Thanks, but this is handled upstream.

alvaro-paco:
May you considerate this. The crash occurs with redux-logger middleware.

